PR #4, opened 2020-05-26
Add Context type

### LINKED ISSUE #2: Build a Context PoC
In its initial version, `Context` will be a wrapper around a dictionary (`[String : Any]`) with methods to `inject` and `extract` metadata.
The shape of both `Context` and its storage will likely change in the future, but this first iteration allows us to start experimenting with potential use-cases.

slashmo:
Adds a `Context` type that uses a type-safe dictionary under the hood to store metadata.

Closes #2


Sources/ContextPropagation/Context.swift
```diff
@@ -0,0 +1,29 @@
+public protocol ContextKey {
+    associatedtype Value
+}
+
+public struct Context {
+    private var dict = [ObjectIdentifier: ValueContainer]()
```
ktoso, line 6: Ok, that will get us going with the PoC 👍 That'll be useful, regardless of what we'll end up storing in those.

Sources/ContextPropagation/Context.swift
```diff
@@ -0,0 +1,29 @@
...
+
+    public init() {}
+
+    public mutating func inject<Key: ContextKey>(_ key: Key.Type, value: Key.Value) {
```
ktoso, line 10: I expect those will change a lot, but let's get there when we get there -- driven by use cases, which we should start writing up next (as tests or mini apps perhaps) to inform the design :-)

Kind of hopping ahead here, but as we talked in our discussion about "where" the logic to inject/extract has to live, so that will inform how these APIs have to look like. If we'll end up with some "Injector" or similar protocols that can "inject into an HTTPRequest" or similar etc.

Sources/ContextPropagation/Context.swift
```diff
@@ -0,0 +1,29 @@
+public protocol ContextKey {
+    associatedtype Value
+}
+
+public struct Context {
```
ktoso, line 5: Do we want to leave out the naming bikeshedding for now or tackle it right away? :-)

(Long story short, I'm a bit worrying about the "catch all" name that context is and keen on phrasing it as baggage since we "carry it around" with inspiration from terminology https://github.com/tracingplane/tracingplane-go/blob/master/tracingplane/baggagecontext.go and the in the paper we've been reviewing and getting inspiration from 🤔) Let's have the naming as a separate topic though, ignore for now! 🙇

ktoso (approved):
Ok, let's get the ball rolling :-)

There will be a lot of churn in this API I expect, so it's important to get something in here, and iterate on it using use cases :-)

Sources/ContextPropagation/Context.swift
slashmo: I agree that `Context` is a bit too generic. I'd say we can leave it like this as long as we are the only users of the API and can simply Refactor>Rename once we found a better name for this informed by the different use cases.

Sources/ContextPropagation/Context.swift
slashmo: Yes, I think we can tackle the "injector" once we have multiple use cases in place to figure out where to best build abstractions.

ktoso: Ok :)

Sources/ContextPropagation/Context.swift
ktoso: I'm ok with that :) Let's revisit at some point

PR #11, opened 2020-06-01
Add usecase target for manual context propagation

### LINKED ISSUE #6: Take the Context for a spin
As our first use case of the context, we'll write some functions inside a Unit Test that take a `Context` and `inject`, `extract` metadata.

slashmo:
Adds a `ManualContextPropagation` use-case target simulating the following actions:

- Handle incoming fake request
- Propagate context
  - Handle request by making subsequent request
  - Populate outgoing request headers with trace ID from `Context`

Closes #6

### DIFF
--- a/Sources/ContextPropagation/InstrumentationMiddleware.swift
+++ b/Sources/ContextPropagation/InstrumentationMiddleware.swift
@@ -1,0 +1,28 @@
+public protocol InstrumentationMiddlewareProtocol {
+    associatedtype ExtractFrom
+    associatedtype InjectInto
+
+    func extract(from: ExtractFrom, into context: inout Context)
+    func inject(from context: Context, into: inout InjectInto)
+}
+
+public struct InstrumentationMiddleware<InjectInto, ExtractFrom>: InstrumentationMiddlewareProtocol {
+    private let extract: (ExtractFrom, inout Context) -> Void
+    private let inject: (Context, inout InjectInto) -> Void
+
+    public init(
+        extract: @escaping (ExtractFrom, inout Context) -> Void,
+        inject: @escaping (Context, inout InjectInto) -> Void
+    ) {
+        self.extract = extract
+        self.inject = inject
+    }
+
+    public func extract(from: ExtractFrom, into context: inout Context) {
+        self.extract(from, &context)
+    }
+
+    public func inject(from context: Context, into: inout InjectInto) {
+        self.inject(context, &into)
+    }
+}
--- a/UseCases/Package.swift
+++ b/UseCases/Package.swift
@@ -1,0 +1,17 @@
+// swift-tools-version:5.2
+import PackageDescription
+
+let package = Package(
+    name: "use-cases",
+    products: [
+        .executable(name: "ManualContextPropagation", targets: ["ManualContextPropagation"])
+    ],
+    dependencies: [
+        .package(path: "../")
+    ],
+    targets: [
+        .target(name: "ManualContextPropagation", dependencies: [
+            .product(name: "ContextPropagation", package: "gsoc-swift-tracing")
+        ])
+    ]
+)
--- a/UseCases/Sources/ManualContextPropagation/main.swift
+++ b/UseCases/Sources/ManualContextPropagation/main.swift
@@ -14,29 +14,6 @@ let server = FakeHTTPServer(
 print("=== Receive HTTP request on server ===")
 server.receive(FakeHTTPRequest(path: "/", headers: []))
 
-// MARK: - Instrumentation Middleware
-
-protocol InstrumentationMiddlewareProtocol {
-    associatedtype ExtractFrom
-    associatedtype InjectInto
-
-    func extract(from: ExtractFrom, into context: inout Context)
-    func inject(from context: Context, into: inout InjectInto)
-}
-
-struct InstrumentationMiddleware<InjectInto, ExtractFrom>: InstrumentationMiddlewareProtocol {
-    let extract: (ExtractFrom, inout Context) -> Void
-    let inject: (Context, inout InjectInto) -> Void
-
-    func extract(from: ExtractFrom, into context: inout Context) {
-        self.extract(from, &context)
-    }
-
-    func inject(from context: Context, into: inout InjectInto) {
-        self.inject(context, &into)
-    }
-}
-
 // MARK: - Fake HTTP Server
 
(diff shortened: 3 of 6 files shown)


UseCases/Package.swift
```diff
@@ -0,0 +1,17 @@
+// swift-tools-version:5.2
+import PackageDescription
+
+let package = Package(
+    name: "use-cases",
+    products: [
+        .executable(name: "ManualContextPropagation", targets: ["ManualContextPropagation"])
+    ],
+    dependencies: [
+        .package(path: "../")
```
ktoso, line 10: 👍

UseCases/Sources/ManualContextPropagation/main.swift
```diff
@@ -0,0 +1,130 @@
...
+}
+
+print("=== Receive HTTP request on server ===")
+server.receive(FakeHTTPRequest(path: "/", headers: []))
+
+// MARK: - Instrumentation Middleware
+
+protocol InstrumentationMiddlewareProtocol {
+    associatedtype ExtractFrom
+    associatedtype InjectInto
```
ktoso: For benefit of other readers: names are subject to change, we're not marrying them just yet 😉 
Those are ok for the time being, if we find better ones we'd change.

UseCases/Sources/ManualContextPropagation/main.swift
```diff
@@ -0,0 +1,130 @@
...
+    let inject: (Context, inout InjectInto) -> Void
+
+    func extract(from: ExtractFrom, into context: inout Context) {
+        self.extract(from, &context)
+    }
+
+    func inject(from context: Context, into: inout InjectInto) {
+        self.inject(context, &into)
+    }
+}
```
ktoso: 👍

UseCases/Sources/ManualContextPropagation/main.swift
```diff
@@ -0,0 +1,130 @@
...
+
+typealias HTTPHeaders = [(String, String)]
```
ktoso, line 19: move under the MARK below? it's part of the fake server after all, right?

UseCases/Sources/ManualContextPropagation/main.swift
```diff
@@ -0,0 +1,130 @@
...
+    client.performRequest(context, request: outgoingRequest)
+    return FakeHTTPResponse()
+}
+
+print("=== Receive HTTP request on server ===")
+server.receive(FakeHTTPRequest(path: "/", headers: []))
+
+// MARK: - Instrumentation Middleware
+
+protocol InstrumentationMiddlewareProtocol {
```
ktoso: Would `InstrumentationMiddlewareProtocol` and its friends want to live in the library, rather than the use case?

UseCases/Sources/ManualContextPropagation/main.swift
```diff
@@ -0,0 +1,130 @@
...
+
+        func extract(from headers: HTTPHeaders, into context: inout Context) {
+            let traceID = headers.first(where: { $0.0 == FakeTraceID.headerName })?.1 ?? tracer.generateTraceID()
+            context.inject(FakeTraceID.self, value: traceID)
+        }
+
+        func inject(from context: Context, into headers: inout HTTPHeaders) {
+            guard let traceID = context.extract(FakeTraceID.self) else { return }
+            headers.append((FakeTraceID.headerName, traceID))
+        }
```
ktoso, line 99: inject/extract look good 👍

UseCases/Sources/ManualContextPropagation/main.swift
```diff
@@ -0,0 +1,130 @@
...
+    struct Middleware: InstrumentationMiddlewareProtocol {
+        private let tracer: FakeTracer
+
+        init(tracer: FakeTracer) {
+            self.tracer = tracer
+        }
+
+        func extract(from headers: HTTPHeaders, into context: inout Context) {
+            let traceID = headers.first(where: { $0.0 == FakeTraceID.headerName })?.1 ?? tracer.generateTraceID()
+            context.inject(FakeTraceID.self, value: traceID)
```
ktoso, line 93: food for thought (no need to act on this now);

it may be useful that we use the words "inject" "extract" only in a single API - what currently is the Middleware (and seems like the right spot for it as well) - and the calls on context could be get/set. The reason for that is when one discusses an API and someone mentions "and i inject..." then we know it's about a middle ware, rather than the "set" action which does not have any logic, just sets a thing.

I think we've discussed that before when we looked at some other context library or am I misremembering that? Not urgent nor blocking this PR but food for thought :)

ktoso (approved):
LGTM, good next step :) 

Next up NIO handlers also as use-case and perhaps making a "common middleware" that we'll use in all our examples? We can then start carving out the boundaries of libraries and where what type would live, wdyt?

UseCases/Sources/ManualContextPropagation/main.swift
slashmo: Yep, that'll be tracked as part of #12.

UseCases/Sources/ManualContextPropagation/main.swift
slashmo: Yep, moved them.

Sources/ContextPropagation/InstrumentationMiddleware.swift
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public func extract(from: ExtractFrom, into context: inout Context) {
+        self.extract(from, &context)
+    }
+
+    public func inject(from context: Context, into: inout InjectInto) {
+        self.inject(context, &into)
+    }
+}
```
ktoso, line 28: kk, good move :) We'll see if needs completely separate or here is fine in the long run, but now definitely good here since all use case will share it 👍

UseCases/Sources/ManualContextPropagation/main.swift
slashmo: This is also tracked in #12.

ktoso: Ah right thx

Sources/ContextPropagation/InstrumentationMiddleware.swift
slashmo: Yes, I also briefly considered created a separate library for that but figured that we'll get around to that eventually once we gathered more insights ✅

PR #13, opened 2020-06-04
Add usecase for manual context propagation with AsyncHTTPClient

### LINKED ISSUE #8: Inject/Extract HTTP Headers
- Manually inject HTTP headers derived from `Context` into `AsyncHTTPClient`
- Manually extract HTTP headers from `AsyncHTTPClient` and store in `Context`

slashmo:
## New use case

Adds a `ManualAsyncHTTPClient` use-case target that uses a `FakeHTTPServer` and `FakeTracer` to simulate the following steps.

- Handle incoming fake request
- Propagate context
  - Handle request by making subsequent request using `AsyncHTTPClient`
  - Populate outgoing request (`HTTPClient.Request`) headers with trace ID from `Context`

## Discussion

The `HTTPClient` integration is done by wrapping the client inside an `InstrumentingHTTPClient` struct that adds an `execute` method similar to the one of `AsyncHTTPClient` additionally receiving the context as an argument.

Wrapping `AsyncHTTPClient` in a struct is only a temporary solution to demonstrate without forking it what it's new, context-aware, API could look like.

Closes #8

### DIFF
--- a/UseCases/Sources/ManualAsyncHTTPClient/main.swift
+++ b/UseCases/Sources/ManualAsyncHTTPClient/main.swift
@@ -10,8 +10,7 @@ let server = FakeHTTPServer(
         url: "https://swift.org",
         headers: ["Accept": "application/json"]
     )
-    client.execute(context, request: outgoingRequest)
-//    client.execute(request: outgoingRequest, inContext: context)
+    client.execute(request: outgoingRequest, context: context)
     return FakeHTTPResponse()
 }
 
@@ -32,15 +31,10 @@ struct InstrumentedHTTPClient {
         }
     }
 
-    func execute(request: HTTPClient.Request, inContext context: Context) {
-        execute(context, request: request)
-    }
-
-    func execute(_ context: Context, request: HTTPClient.Request) {
+    func execute(request: HTTPClient.Request, context: Context) {
         var request = request
         instrumentationMiddlewares.forEach { $0.inject(from: context, into: &request.headers) }
         print(request.headers)
-//        client.execute(request: request)
     }
 }
 
--- a/UseCases/Sources/ManualAsyncHTTPClient/main.swift
+++ b/UseCases/Sources/ManualAsyncHTTPClient/main.swift
@@ -18,9 +18,9 @@ let server = FakeHTTPServer(
 print("=== Receive HTTP request on server ===")
 server.receive(try! HTTPClient.Request(url: "https://swift.org"))
 
-// MARK: - InstrumentingHTTPClient
+// MARK: - InstrumentedHTTPClient
 
-struct InstrumentingHTTPClient {
+struct InstrumentedHTTPClient {
     private let client = HTTPClient(eventLoopGroupProvider: .createNew)
     private let instrumentationMiddlewares: [InstrumentationMiddleware<HTTPHeaders, HTTPHeaders>]
 
@@ -51,11 +51,11 @@ struct FakeHTTPResponse {}
 private typealias HTTPHeadersInstrumentationMiddleware = InstrumentationMiddleware<HTTPHeaders, HTTPHeaders>
 
 struct FakeHTTPServer {
-    typealias Handler = (Context, HTTPClient.Request, InstrumentingHTTPClient) -> FakeHTTPResponse
+    typealias Handler = (Context, HTTPClient.Request, InstrumentedHTTPClient) -> FakeHTTPResponse
 
     private let instrumentationMiddlewares: [InstrumentationMiddleware<HTTPHeaders, HTTPHeaders>]
     private let catchAllHandler: Handler
-    private let client: InstrumentingHTTPClient
+    private let client: InstrumentedHTTPClient
 
     init<M: InstrumentationMiddlewareProtocol>(
         instrumentationMiddlewares: [M],
@@ -64,7 +64,7 @@ struct FakeHTTPServer {
         self.instrumentationMiddlewares = instrumentationMiddlewares.map {
             InstrumentationMiddleware(extract: $0.extract, inject: $0.inject)
         }
-        self.client = InstrumentingHTTPClient(instrumentationMiddlewares: instrumentationMiddlewares)
+        self.client = InstrumentedHTTPClient(instrumentationMiddlewares: instrumentationMiddlewares)
         self.catchAllHandler = catchAllHandler
     }
 
--- a/UseCases/Sources/ManualAsyncHTTPClient/main.swift
+++ b/UseCases/Sources/ManualAsyncHTTPClient/main.swift
@@ -1,61 +1,27 @@
+import AsyncHTTPClient
 import ContextPropagation
-
-// MARK: - Demo
+import NIOHTTP1
 
 let server = FakeHTTPServer(
     instrumentationMiddlewares: [FakeTracer.Middleware(tracer: FakeTracer())]
 ) { context, request, client -> FakeHTTPResponse in
     print("=== Perform subsequent request ===")
-    let outgoingRequest = FakeHTTPRequest(path: "/other-service", headers: [("Content-Type", "application/json")])
-    client.performRequest(context, request: outgoingRequest)
+    let outgoingRequest = try! HTTPClient.Request(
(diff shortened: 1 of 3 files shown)


slashmo:
Please note that all naming is totally undecided yet, but I think we can already start a discussion about where to put the context argument in `AsyncHTTPClient`s methods. The example shows two methods, one of which puts the context first and one which puts the context last.

1. `client.execute(_:Context, request:HTTPClient.Request)`
2. `client.execute(request:HTTPClient.Request, inContext:Context)`

In my opinion the latter feels more natural in Swift-land.

@weissi & @ktoso I'd love to hear your thoughts on that 💭

ktoso:
Yup, so I'm in the middle of "making sure people are on board with context-last" and will loop back once we have confirmation from some other folks, but in general I'm growing fond of "last (but before closures)"  as well 👍 

Then there's another bikeshed about how the param will look like... 

> `client.execute(request: HTTPClient.Request, inContext: Context)`

So the `inContext` is rather not a good parameter name; since we repeat the type, so if we went with "in" it'd be `in context: Context` but that's to be debated if overloading `in` to always mean context is good... Options here are `_ context: Context`, `with context: ... `, `in context: ...` etc. I'm not really sure either of those are great.

As for "where" I think we'll end up with the "last" but will confirm after some review with folks who design APIs over here so we're "all" on the same page :) 

To be honest I'd love us to revisit the name "context" as well eventually since it's sooooo generic and also overloaded with tons of frameworks which also have their `context`, so often you'd end up with `context.context` if a framework context contains "our" context. So I'd consider adopting the terminology that Pivot Tracing / Tracingplane are pushing since a few years which is `Baggage` or `BaggageContext`; then it's clear that `context` libraries are free to use (and do so already) and "carry around tracing values" is baggage. Not sure what the best time is for this discussion though. I think it's important and there is enough prior art for "baggage"; `client.execute(request, with: baggage)` 😉 

In any case, going to review the PR now and trying to not get too hung up on names yet :-)

UseCases/Sources/ManualAsyncHTTPClient/main.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    init<M: InstrumentationMiddlewareProtocol>(
+        instrumentationMiddlewares: [M]
+    ) where M.InjectInto == HTTPHeaders, M.ExtractFrom == HTTPHeaders {
+        self.instrumentationMiddlewares = instrumentationMiddlewares.map {
+            InstrumentationMiddleware(extract: $0.extract, inject: $0.inject)
+        }
+    }
+
+    func execute(request: HTTPClient.Request, inContext context: Context) {
```
ktoso: if we wanted to adopt this style then it should rather be `in context` or `with context`, or `_ context` or `context: ...`...  (Not too hung up on names so far.)

This is a good chance to point people at how the two things read, thanks for pinging Johannes :) 

// more thoughts: https://github.com/slashmo/gsoc-swift-tracing/pull/13#issuecomment-638755592

UseCases/Sources/ManualAsyncHTTPClient/main.swift
```diff
@@ -0,0 +1,109 @@
...
+//    client.execute(request: outgoingRequest, inContext: context)
+    return FakeHTTPResponse()
+}
+
+print("=== Receive HTTP request on server ===")
+server.receive(try! HTTPClient.Request(url: "https://swift.org"))
+
+// MARK: - InstrumentingHTTPClient
+
+struct InstrumentingHTTPClient {
```
ktoso: `InstrumentedHTTPClient`?

slashmo: 👍

UseCases/Sources/ManualAsyncHTTPClient/main.swift
```diff
@@ -0,0 +1,109 @@
...
+    private let client = HTTPClient(eventLoopGroupProvider: .createNew)
+    private let instrumentationMiddlewares: [InstrumentationMiddleware<HTTPHeaders, HTTPHeaders>]
```
ktoso, line 24: Is the  `[]` getting a bit annoying perhaps?

Should we write an `InstrumentationMiddleware` which is an aggregate or `MultiplexingInstrumentationMiddleware` that can store the array and in all cutpoints we simply call "the middleware" (name to be bikeshed) rather than having to implement the array traversal many times?

You can have a look at https://github.com/apple/swift-log/blob/87c2553da204fc269be89aec39e38217d65c4980/Sources/Logging/Logging.swift#L459-L520 to see how we do it in logging

slashmo: Sounds good! 👍 I like the multiplex approach, will take a look. I think that's also not blocked by the naming as we'll always have multiple "things" to be called.

slashmo:
> rather not a good parameter name; since we repeat the type

`HTTPClient`s current API reads `client.execute(request: request)` instead of `client.execute(request)`, so maybe something simple as `client.execute(request: request, context: context)` would be a good solution.

> but that's to be debated if overloading `in` to always mean context is good...

Agreed, I think that `in` or `with` are a bit too generic and already used in other contexts (no pun intended).

> So I'd consider adopting the terminology that Pivot Tracing / Tracingplane are pushing since a few years which is `Baggage` or `BaggageContext`;

`client.execute(request, carrying: baggage)` came to my mind while reading this.

UseCases/Sources/ManualAsyncHTTPClient/main.swift
slashmo: Commented on in https://github.com/slashmo/gsoc-swift-tracing/pull/13#issuecomment-638766188

weissi:
- +1 on it should be `client.execute(request)`, `request:` is superfluous. If we feel we need it it should be `executeRequest(request)` but that ship has sailed
- +1 on if possible avoiding the word "context" in general, it doesn't mean very much at all. I like `BaggageContext` or `Baggage` a little better but probably not enough. It still doesn't mean much. I do like SwiftUI's choice of `Environment`, I think it gives people the right idea.

@ktoso (ignoring the debate on which word is right): regarding `context:` being duplicated: I don't think that's bad in this case. `Context` is pretty special in that people should ignore it when reading the code really. And `context:` tells them exactly what this is and kind of acts as a marker for "yeah, I'll ignore this, this is the context/environment/baggage that I'm just carrying around".

Same for `logger:` really. I think argument labels can also be used as markers for "this isn't that important for the methods functionality but for technical reasons [debugging/tracing] we need it"

ktoso:
> @ktoso (ignoring the debate on which word is right): regarding context: being duplicated: I don't think that's bad in this case. Context is pretty special in that people should ignore it when reading the code really. And context: tells them exactly what this is and kind of acts as a marker for "yeah, I'll ignore this, this is the context/environment/baggage that I'm just carrying around".

That's a good point really; realistically noone would really "read out loud" the context/baggage (or logger for that matter), so I agree that trying to make it "read smooth" with in/with etc naming is likely not helping (going against that "don't read it out loud") -- that's a good thought here :)

UseCases/Sources/ManualAsyncHTTPClient/main.swift
ktoso: Yup, just something to do separately at some point -- just because  the `[]` passing around gets annoying :) 
Maybe a mini ticket?

ktoso (approved):

UseCases/Sources/ManualAsyncHTTPClient/main.swift
slashmo: Reported in #14

slashmo:
> That's a good point really; realistically noone would really "read out loud" the context/baggage (or logger for that matter), so I agree that trying to make it "read smooth" with in/with etc naming is likely not helping (going against that "don't read it out loud") -- that's a good thought here :)

That's a good point indeed 👍 I updated the example and will merge now. I have a feeling we'll iterate a lot on the naming (especially `Context`/`InstrumentationMiddleware`, but I think not having it "read smooth" is the best thing to do, no matter what we end calling it.

PR #16, opened 2020-06-05
Add MultiplexInstrumentationMiddleware

### LINKED ISSUE #14: Create MultiplexMiddleware
Currently, it's annoying for each instrumented system to deal with multiple instrumentation middleware. Instead, we want to create a `MultiplexInstrumentationMiddleware` that takes all other middleware on init and itself conforms to `InstrumentationMiddlewareProtocol`. Then instrumented systems can think of the whole instrumentation stack as one middleware that under the hood calls out to all middleware.

Prior art for this is [Logging 's `MultiplexLogHandler`](https://github.com/apple/swift-log/blob/master/Sources/Logging/Logging.swift#L464).

slashmo:
Closes #14

### DIFF
--- a/Sources/ContextPropagation/Instrumentation/InstrumentationMiddleware.swift
+++ b/Sources/ContextPropagation/Instrumentation/InstrumentationMiddleware.swift
@@ -10,12 +10,13 @@ public struct InstrumentationMiddleware<InjectInto, ExtractFrom>: Instrumentatio
     private let extract: (ExtractFrom, inout Context) -> Void
     private let inject: (Context, inout InjectInto) -> Void
 
-    public init(
-        extract: @escaping (ExtractFrom, inout Context) -> Void,
-        inject: @escaping (Context, inout InjectInto) -> Void
-    ) {
-        self.extract = extract
-        self.inject = inject
+    public init<Middleware>(_ middleware: Middleware)
+        where
+        Middleware: InstrumentationMiddlewareProtocol,
+        Middleware.InjectInto == InjectInto,
+        Middleware.ExtractFrom == ExtractFrom {
+            self.extract = middleware.extract
+            self.inject = middleware.inject
     }
 
     public func extract(from: ExtractFrom, into context: inout Context) {
--- a/Sources/ContextPropagation/Instrumentation/MultiplexInstrumentationMiddleware.swift
+++ b/Sources/ContextPropagation/Instrumentation/MultiplexInstrumentationMiddleware.swift
@@ -1,0 +1,23 @@
+/// A pseudo-`InstrumentationMiddlewareProtocol` that can be used to instrument using multiple other `InstrumentationMiddleware`s across
+/// a common `Context`.
+public struct MultiplexInstrumentationMiddleware<InjectInto, ExtractFrom> {
+    private var middlewares: [InstrumentationMiddleware<InjectInto, ExtractFrom>]
+
+    /// Create a `MultiplexInstrumentationMiddleware`.
+    ///
+    /// - Parameter middlewares: An array of `InstrumentationMiddleware`s, each of which will be used to `extract` from and `inject` into
+    /// a common `Context`.
+    public init(_ middlewares: [InstrumentationMiddleware<InjectInto, ExtractFrom>]) {
+        self.middlewares = middlewares
+    }
+}
+
+extension MultiplexInstrumentationMiddleware: InstrumentationMiddlewareProtocol {
+    public func extract(from: ExtractFrom, into context: inout Context) {
+        middlewares.forEach { $0.extract(from: from, into: &context) }
+    }
+
+    public func inject(from context: Context, into: inout InjectInto) {
+        middlewares.forEach { $0.inject(from: context, into: &into) }
+    }
+}
--- a/Tests/ContextPropagationTests/Instrumentation/InstrumentationMiddlewareTests.swift
+++ b/Tests/ContextPropagationTests/Instrumentation/InstrumentationMiddlewareTests.swift
@@ -1,0 +1,68 @@
+import ContextPropagation
+import XCTest
+
+final class InstrumentationMiddlewareTests: XCTestCase {
+    func testMultiplexInvokesAllMiddleware() {
+        let middleware = MultiplexInstrumentationMiddleware([
+            InstrumentationMiddleware(FirstFakeTracer.Middleware()),
+            InstrumentationMiddleware(SecondFakeTracer.Middleware())
+        ])
+
+        var context = Context()
+        let requestHeaders = [String: String]()
+        middleware.extract(from: requestHeaders, into: &context)
+
+        XCTAssertEqual(context.extract(FirstFakeTracer.TraceID.self), FirstFakeTracer.defaultTraceID)
+        XCTAssertEqual(context.extract(SecondFakeTracer.TraceID.self), SecondFakeTracer.defaultTraceID)
+
+        var subsequentRequestHeaders = [String: String]()
+        middleware.inject(from: context, into: &subsequentRequestHeaders)
+
+        XCTAssertEqual(subsequentRequestHeaders, [
+            FirstFakeTracer.headerName: FirstFakeTracer.defaultTraceID,
+            SecondFakeTracer.headerName: SecondFakeTracer.defaultTraceID
+        ])
+    }
+}
+
+private struct FirstFakeTracer {
+    enum TraceID: ContextKey {
+        typealias Value = String
+    }
+
+    static let headerName = "first-fake-trace-id"
+    static let defaultTraceID = UUID().uuidString
+
+    struct Middleware: InstrumentationMiddlewareProtocol {
+
+        func extract(from headers: [String: String], into context: inout Context) {
+            let traceID = headers.first(where: { $0.key == FirstFakeTracer.headerName })?.value ?? FirstFakeTracer.defaultTraceID
+            context.inject(FirstFakeTracer.TraceID.self, value: traceID)
+        }
+
+        func inject(from context: Context, into headers: inout [String: String]) {
+            headers[FirstFakeTracer.headerName] = context.extract(FirstFakeTracer.TraceID.self)
+        }
+    }
+}
+
+private struct SecondFakeTracer {
+    enum TraceID: ContextKey {
+        typealias Value = String
+    }
+
+    static let headerName = "second-fake-trace-id"
+    static let defaultTraceID = UUID().uuidString
+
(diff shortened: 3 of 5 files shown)


slashmo:
@ktoso This is a first PoC of a `MultiplexInstrumentationMiddleware`. Unfortunately, it involves some type meddling because of the associated types. I'm especially unhappy about `MultiplexInstrumentationMiddleware`s initializer, where every `InstrumentationMiddlewareProtocol` conforming type needs to be wrapped in an `InstrumentationMiddleware`. Do you know of a better way to handle this?

*So far I only used the `MultiplexInstrumentationMiddleware` in the manual propagation use case (without `AsyncHTTPClient`), because I wanted to see if we can minimize the type meddling first.*

ktoso:
> I'm especially unhappy about MultiplexInstrumentationMiddlewares initializer, where every InstrumentationMiddlewareProtocol conforming type needs to be wrapped in an InstrumentationMiddleware. Do you know of a better way to handle this?

Yep, that's (for better or worse) what we'll need to do here indeed. This happens everywhere where one might want to "store" a value of a protocol with associated types. So indeed, we have do to this map there and there's no way around it today.

This will be solved once Swift would gets a few features, including "generalized existentials" (that's the exact missing type system feature we're hitting here): https://github.com/austinzheng/swift-evolution/blob/az-existentials/proposals/XXXX-enhanced-existentials.md and some syntax to express those like Joe proposes here https://forums.swift.org/t/protocol-assoctype-t-shorthand-for-combined-protocol-and-associated-type-constraints-without-naming-the-constrained-type/21217 all this is part of a "grand scheme of things" future direction for generics explained here https://forums.swift.org/t/improving-the-ui-of-generics/22814

but yeah, today we do need that wrapping and iterating over all the values indeed.

Sources/ContextPropagation/Instrumentation/InstrumentationMiddleware.swift
```diff
@@ -18,6 +18,10 @@ public struct InstrumentationMiddleware<InjectInto, ExtractFrom>: Instrumentatio
         self.inject = inject
     }
 
+    public init<M: InstrumentationMiddlewareProtocol>(_ middleware: M) where M.InjectInto == InjectInto, M.ExtractFrom == ExtractFrom {
```
ktoso: That's a super long line, can we cut it down like so perhaps:

```suggestion
public init<Middleware>(_ middleware: Middleware)
        where
        Middleware: InstrumentationMiddlewareProtocol,
        Middleware.ExtractFrom == Self.ExtractFrom,
        Middleware.InjectInto == Self.InjectInto {
```

Sources/ContextPropagation/Instrumentation/MultiplexInstrumentationMiddleware.swift
```diff
@@ -0,0 +1,21 @@
+public struct MultiplexInstrumentationMiddleware<InjectInto, ExtractFrom> {
+    private var middleware: [InstrumentationMiddleware<InjectInto, ExtractFrom>]
```
ktoso:
```suggestion
    private var middlewares: [InstrumentationMiddleware<InjectInto, ExtractFrom>]
```

?

Sources/ContextPropagation/Instrumentation/MultiplexInstrumentationMiddleware.swift
```diff
@@ -0,0 +1,21 @@
...
+
+//    public init(_ middleware: [InstrumentationMiddleware<InjectInto, ExtractFrom>]) {
+//        self.middleware = middleware
+//    }
+
+    public init<M: InstrumentationMiddlewareProtocol>(_ middleware: [M]) where M.InjectInto == InjectInto, M.ExtractFrom == ExtractFrom {
+        self.middleware = middleware.map(InstrumentationMiddleware.init)
```
ktoso: yep, we'll indeed need this (explained more in top level comment on the PR)

Tests/ContextPropagationTests/Instrumentation/InstrumentationMiddlewareTests.swift
```diff
@@ -0,0 +1,68 @@
...
+        XCTAssertEqual(context.extract(SecondFakeTracer.TraceID.self), SecondFakeTracer.defaultTraceID)
+
+        var subsequentRequestHeaders = [String: String]()
+        middleware.inject(from: context, into: &subsequentRequestHeaders)
+
+        XCTAssertEqual(subsequentRequestHeaders, [
+            FirstFakeTracer.headerName: FirstFakeTracer.defaultTraceID,
+            SecondFakeTracer.headerName: SecondFakeTracer.defaultTraceID
+        ])
+    }
```
ktoso, line 25: 👍

UseCases/Sources/ManualContextPropagation/main.swift
```diff
@@ -2,9 +2,10 @@ import ContextPropagation
 
 // MARK: - Demo
 
-let server = FakeHTTPServer(
-    instrumentationMiddlewares: [FakeTracer.Middleware(tracer: FakeTracer())]
-) { context, request, client -> FakeHTTPResponse in
+let server = FakeHTTPServer(instrumentationMiddleware: MultiplexInstrumentationMiddleware([
+    InstrumentationMiddleware(FakeTracer.Middleware(tracer: FakeTracer())),
+    InstrumentationMiddleware(SecondFakeTracer.Middleware(tracer: SecondFakeTracer()))
+])) { context, request, client -> FakeHTTPResponse in
```
ktoso: maybe no need to show the first/second stuff in the samples? We have the test for that, we know it works :)

UseCases/Sources/ManualContextPropagation/main.swift
```diff
@@ -105,3 +102,33 @@ private struct FakeTracer {
         static let headerName = "fake-trace-id"
     }
 }
+
+private struct SecondFakeTracer {
```
ktoso: maybe no need to showcase these in the samples? will add noise and we know how the multiplex one works from the tests (and docs)

Sources/ContextPropagation/Instrumentation/MultiplexInstrumentationMiddleware.swift
```diff
@@ -0,0 +1,21 @@
+public struct MultiplexInstrumentationMiddleware<InjectInto, ExtractFrom> {
```
ktoso, line 3: Add some documentation?

You can base it on https://github.com/apple/swift-log/blob/master/Sources/Logging/Logging.swift#L459-L464

basically replacing the "logging" words with our domain :)

Sources/ContextPropagation/Instrumentation/InstrumentationMiddleware.swift
```diff
@@ -18,6 +18,10 @@ public struct InstrumentationMiddleware<InjectInto, ExtractFrom>: Instrumentatio
...
+        self = .init(extract: middleware.extract, inject: middleware.inject)
```
ktoso: This init is a bit meh, we'd force everyone who has to perform the wrapping to do the "give me the callbacks" manually which gets annoying, consider the following init implementation:

```swift
public struct InstrumentationMiddleware<InjectInto, ExtractFrom>: InstrumentationMiddlewareProtocol {
    private let _extract: (ExtractFrom, inout Context) -> Void
    private let _inject: (Context, inout InjectInto) -> Void

    public init<Middleware>(_ middleware: Middleware)
        where
        Middleware: InstrumentationMiddlewareProtocol,
        Middleware.ExtractFrom == Self.ExtractFrom,
        Middleware.InjectInto == Self.InjectInto {
            _extract = { from, context in middleware.extract(from: from, into: &context) }
            _inject = { context, into in middleware.inject(from: context, into: &into) }
    }

    public func extract(from: ExtractFrom, into context: inout Context) {
        _extract(from, &context)
    }

    public func inject(from context: Context, into: inout InjectInto) {
        _inject(context, &into)
    }
}

```

Then the callsite here can be

```suggestion
        self = .init(middleware)
```

---

we could also avoid "double wrapping" with

```
        if let alreadyWrapped = middleware as? InstrumentationMiddleware<InjectInto, ExtractFrom> {
            self = alreadyWrapped
        } else {
...
```

or making the wrapping a function on the protocol `as...: InstrumentationMiddleware { return self }` (if already this), and `: InstrumentationMiddleware { .init(self) }` by default on the protocol... 

But that's not really necessary right now; though let's change the init so it takes a middleware and not the functions though 👍

Sources/ContextPropagation/Instrumentation/MultiplexInstrumentationMiddleware.swift
```diff
@@ -0,0 +1,21 @@
...
+    private var middleware: [InstrumentationMiddleware<InjectInto, ExtractFrom>]
+
+//    public init(_ middleware: [InstrumentationMiddleware<InjectInto, ExtractFrom>]) {
+//        self.middleware = middleware
+//    }
+
+    public init<M: InstrumentationMiddlewareProtocol>(_ middleware: [M]) where M.InjectInto == InjectInto, M.ExtractFrom == ExtractFrom {
```
ktoso: I think I'd suggest doing the following style everywhere for such long where clauses:

```suggestion
    public init<Middleware>(_ middlewares: [Middleware])
        where
        Middleware: InstrumentationMiddlewareProtocol,
        Middleware.ExtractFrom == Self.ExtractFrom,
        Middleware.InjectInto == Self.InjectInto {
```

it reads much better 😄

Sources/ContextPropagation/Instrumentation/MultiplexInstrumentationMiddleware.swift
```diff
@@ -0,0 +1,21 @@
+public struct MultiplexInstrumentationMiddleware<InjectInto, ExtractFrom> {
+    private var middleware: [InstrumentationMiddleware<InjectInto, ExtractFrom>]
+
+//    public init(_ middleware: [InstrumentationMiddleware<InjectInto, ExtractFrom>]) {
+//        self.middleware = middleware
+//    }
```
ktoso: (dont forget to remove)

ktoso (approved):
Some feedbacks in line; thanks for doing a test for the multiplex handler 👍 

After addressing the minor thingies and perhaps not doing the multiplex handler in the use-cases at all (they distract a bit from the core of what the samples are?) please feel free to merge 👍

ktoso:
(Mini organizational note: if a PR has an issue it resolves, and the issue already has a milestone, let's not assign the PR also to the milestone -- by looking at a milestone we can easily skim "what was one" so there's no need to have "one thing that was done" show up twice :-))

Issue #17, opened 2020-06-08
Add swiftformat and "sanity" script


ktoso:
Similar to other SSWG projects, so we can avoid formatting conflicts :)

No strong opinions on the rules, I'll copy what I have but feel free to put other rules in the config if you have strong opinions on something.

slashmo:
Thanks, closed by #18

ktoso:
Thanks for bringing it up :)

PR #18, opened 2020-06-08
+sanity #17 add sanity scripts (formatting, license)

ktoso:
Simple formatting hook, similar to the one used in NIO and other projects.

I didn't add license validation yet. We'd eventually want the same "header" as we do in NIO and others: https://github.com/apple/swift-nio/blob/master/Tests/NIOTestUtilsTests/ByteToMessageDecoderVerifierTest%2BXCTest.swift#L1-L13 but didn't add it yet.

The settings inspired by NIO and some other projects where we use swiftformat, hope they're fine @slashmo 

// we're sticking to `SwiftFormat, version 0.44.6` in our builds nowadays AFAICS (docker images etc), so that's what this was formatted with.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,7 +1,8 @@
 # file options
 
---swiftversion 5.1
+--swiftversion 5.2
 --exclude .build
+--exclude UseCases/.build
 
 # format options
 
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,0 +1,16 @@
+# file options
+
+--swiftversion 5.1
+--exclude .build
+
+# format options
+
+--ifdef no-indent
+--patternlet inline
+--self insert
+--stripunusedargs closure-only
+--wraparguments before-first
+
+# rules
+
+--disable blankLinesAroundMark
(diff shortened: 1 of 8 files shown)


.swiftformat
```diff
@@ -0,0 +1,16 @@
+# file options
+
+--swiftversion 5.1
```
slashmo: Do we want to use 5.2 here as that’s the version we use in `Package.swift` as well?

.swiftformat
```diff
@@ -0,0 +1,16 @@
...
+--exclude .build
```
slashmo, line 4: Would this also exclude the `.build` folder in `UseCases`?

slashmo (changes requested):
@ktoso Thanks for adding this! 👌 Could you please take a look at the inline comments?

.swiftformat
ktoso: Good catch, I think it wouldnt -- will add

.swiftformat
ktoso: lemme check, can't remember why we still have 5.1 here even though our projects are 5.2 hm :)

slashmo (approved):

PR #19, opened 2020-06-08
Compile UseCases on CI

### LINKED ISSUE #15: Compile use cases on CI
Now that the number of use cases is growing we might want to start adding them to the CI pipeline. @ktoso Do you think this makes sense? Or should we rather write separate integration tests?

slashmo:
As discussed in #15, the `UseCases` package will now be compiled on the CI.

Closes #15

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,3 +11,8 @@ jobs:
         uses: actions/checkout@v1
       - name: Build & Test
         run: swift test -c release --enable-test-discovery
+      - name: Compile UseCases
+        run: |
+          cd UseCases
+          swift package resolve
+          swift build


.github/workflows/ci.yaml
```diff
@@ -11,3 +11,8 @@ jobs:
         uses: actions/checkout@v1
       - name: Build & Test
         run: swift test -c release --enable-test-discovery
+      - name: Compile UseCases
+        run: |
+          cd UseCases
+          swift package resolve
```
ktoso, line 17: Not sure the explicit resolve is needed?

ktoso (approved):
LGTM :)

.github/workflows/ci.yaml
slashmo: `swift build` will do it automatically, but on CI I'd rather be explicit so that it's easier to find failures once they occur. And who knows, maybe some day we'll be able to even cache SPM dependencies 😁

PR #20, opened 2020-06-08
Add NIO HTTP Channel handlers for context/header propagation

### LINKED ISSUE #9: Create NIO outbound handler for injecting HTTP Headers
This handler will contain a `Context` from which it derives HTTP headers to be set on an HTTP response.

### LINKED ISSUE #10: Create NIO inbound handler for extracting HTTP Headers
This handler will contain a `Context` which it populates with trace information extracted from HTTP request headers.

slashmo:
This adds a new library target `NIOInstrumentation` (not happy about that name). It contains:
- `ContextInboundHTTPHandler` (Receives HTTP requests and uses its `instrumentationMiddleware` to propagate the `Context` based on request headers)
- `ContextOutboundHTTPHandler` (Sends HTTP requests and uses its `instrumentationMiddleware` to propagate the request headers based on the `Context`)

Because we currently have no way of passing through `Context` in NIO, I created a `Context.shared` singleton. This of course needs to change soon, but works for now.

I also updated the CI config to parallelize compilation of use-cases and execution of unit tests to speed it up a bit.

Closes #9, closes #10


.github/workflows/ci.yaml
```diff
@@ -3,16 +3,23 @@ name: CI
...
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Resolve Swift dependencies
+        run: cd UseCases && swift package resolve
       - name: Compile UseCases
-        run: |
-          cd UseCases
-          swift package resolve
-          swift build
+        run: cd UseCases && swift build
```
ktoso, line 25: 👍

Sources/NIOInstrumentation/ContextOutboundHTTPHandler.swift
```diff
@@ -0,0 +1,24 @@
+import ContextPropagation
+import NIO
+import NIOHTTP1
+
+public final class ContextOutboundHTTPHandler: ChannelOutboundHandler {
+    public typealias OutboundIn = HTTPClientRequestPart
```
ktoso: Let's change this to an `HTTPClientRequestPartWithBaggage` that's just a glorified tuple like we discussed today :) 
This way the write can offer us the accompanying metadata.

Sources/NIOInstrumentation/ContextOutboundHTTPHandler.swift
```diff
@@ -0,0 +1,24 @@
...
+    public init<Middleware>(instrumentationMiddleware: Middleware)
+        where
+        Middleware: InstrumentationMiddlewareProtocol,
+        Middleware.InjectInto == HTTPHeaders,
+        Middleware.ExtractFrom == HTTPHeaders {
+        self.instrumentationMiddleware = InstrumentationMiddleware(instrumentationMiddleware)
+    }
+
+    public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+        guard case .head(var head) = unwrapOutboundIn(data) else { return }
```
ktoso: Thanks to the `HTTPClientRequestPartWithBaggage` this will be that wrapper; so we'll first extract that, then get the .head out of the `HTTPClientRequestPart` part of that, and we'll get the baggage out of it as well.
Then we'll invoke the instrument like `self.instrument.inject(from: baggage, into: &head.headers)`

We can also leave a `// TODO: context.baggage = baggage // consider how we'll offer this capability in NIO` here just before the context.write.

Sources/NIOInstrumentation/Context+Shared.swift
```diff
@@ -0,0 +1,6 @@
+import ContextPropagation
+
+extension Context {
+    /// Singleton `Context` instance to be used in `NIO` until we can pass around `Context` through `NIO`.
+    public static var shared = Context()
```
ktoso: // I know this is just for the PoCing out things but...
Let's not do this, it's terrifying 😄😉 (Due to concurrent access to a global static var).

I'll comment below what we can do instead for now

Sources/NIOInstrumentation/ContextInboundHTTPHandler.swift
```diff
@@ -0,0 +1,23 @@
...
+        where
+        Middleware: InstrumentationMiddlewareProtocol,
+        Middleware.InjectInto == HTTPHeaders,
+        Middleware.ExtractFrom == HTTPHeaders {
+        self.instrumentationMiddleware = InstrumentationMiddleware(instrumentationMiddleware)
+    }
+
+    public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        guard case .head(let head) = unwrapInboundIn(data) else { return }
+        self.instrumentationMiddleware.extract(from: head.headers, into: &.shared)
```
ktoso: I know the .shared is only for showing off we have read the value here, but it's super scary to access a global var like that :)

In this example I think it'll be good enough if we:
- print the context we extracted, and
- leave a `// TODO context.baggage = the extracted baggage // figure out with NIO how we'll want to implement this

Sources/NIOInstrumentation/ContextOutboundHTTPHandler.swift
ktoso: I know that makes testing hard now... but I guess we're testing here rather that "the instrument was invoked" rather than "what exactly the instrument did" so maybe a test instrument can be good that just verifies we've been called if we want a simple test for now?

Tests/NIOInstrumentationTests/ContextInboundHTTPHandlerTests.swift
```diff
@@ -0,0 +1,26 @@
+import ContextPropagation
+import NIO
+import NIOHTTP1
+import NIOInstrumentation
+import XCTest
+
+final class ContextInboundHTTPHandlerTests: XCTestCase {
+    override class func setUp() {
+        super.setUp()
+        Context.shared = Context()
```
ktoso: (see notes in the handlers and the .shared about this)

Tests/NIOInstrumentationTests/FakeTracer.swift
```diff
@@ -0,0 +1,22 @@
+import ContextPropagation
+import NIOHTTP1
+
+struct FakeTracer {
```
ktoso, line 4: 👍 nice to have it here

ktoso (commented):
LGTM though let's avoid the `.shared` trick, it works fince on the embedded loops but in general it'd be pretty scary so let's not keep it around (to not give people the wrong idea when they look at the repo this week)

Tests/NIOInstrumentationTests/FakeTracer.swift
slashmo: Agreed 😜

Sources/NIOInstrumentation/ContextInboundHTTPHandler.swift
```diff
@@ -0,0 +1,27 @@
...
+        Middleware.ExtractFrom == HTTPHeaders {
+        self.instrumentationMiddleware = InstrumentationMiddleware(instrumentationMiddleware)
+        self.onContext = onContext
+    }
+
+    public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        guard case .head(let head) = unwrapInboundIn(data) else { return }
+        var context = Context()
+        self.instrumentationMiddleware.extract(from: head.headers, into: &context)
+        self.onContext?(context)
```
ktoso: 👍

Sources/NIOInstrumentation/ContextInboundHTTPHandler.swift
```diff
@@ -0,0 +1,27 @@
+import ContextPropagation
+import NIO
+import NIOHTTP1
+
+public final class ContextInboundHTTPHandler: ChannelInboundHandler {
+    public typealias InboundIn = HTTPServerRequestPart
+    public typealias InboundOut = HTTPServerRequestPart
+
+    private let instrumentationMiddleware: InstrumentationMiddleware<HTTPHeaders, HTTPHeaders>
+    private var onContext: ((Context) -> Void)?
```
ktoso: Minor: since the onContext init parameter is not optional this does not have to be either

Sources/NIOInstrumentation/ContextOutboundHTTPHandler.swift
```diff
@@ -0,0 +1,27 @@
...
+        Middleware.ExtractFrom == HTTPHeaders {
+        self.instrumentationMiddleware = InstrumentationMiddleware(instrumentationMiddleware)
+    }
+
+    public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+        let requestPartWithBaggage = unwrapOutboundIn(data)
+        guard case .head(var head) = requestPartWithBaggage.requestPart else { return }
+        self.instrumentationMiddleware.inject(from: requestPartWithBaggage.context, into: &head.headers)
+        // TODO: context.baggage = baggage
+        // consider how we'll offer this capability in NIO
```
ktoso, line 24: 👍

Sources/NIOInstrumentation/ContextOutboundHTTPHandler.swift
```diff
@@ -0,0 +1,27 @@
+import ContextPropagation
+import NIO
+import NIOHTTP1
+
+public final class ContextOutboundHTTPHandler: ChannelOutboundHandler {
+    public typealias OutboundIn = HTTPClientRequestPartWithContext
```
ktoso, line 6: 👍

ktoso (approved):
LGTM, minor nitpick :)

PR #21, opened 2020-06-10
add LICENSE.txt

ktoso:
To clarify this is open source / apache licensed and no issues about using it in other projects.

Are you okey with Apache v2 @slashmo? Most SSWG projects use that one and it's fairly simple.

Didn't do license headers for files yet, can do that next (with a script in sanity.sh too)

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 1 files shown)


slashmo:
Thanks for adding it. I’m fine with Apache 2. Should I fill in my name in the blanks?

ktoso:
Nope no need to do that here. 

We'll have to do on source files, it's used like this:

- just the license file: https://github.com/apple/swift-nio/blob/master/LICENSE.txt (same file as we have here now)
- and all sources then use a header like https://github.com/apple/swift-nio/blob/master/Sources/NIO/ByteBuffer-aux.swift#L1-L13 there we'll want to write "who owns this" etc, usually the way NIO does it is quite good, directing at the CONTRIBUTORS.txt file, which then is generated from git logs

Once we have this I can follow up with the license checker script that checks all files have the license header etc

slashmo:
Okay, didn’t know that. Makes sense 👍🏻

slashmo (approved):

PR #34, opened 2020-06-16
Add BaggageLogging library 📦

### LINKED ISSUE #7: Expose Context metadata as Logging metadata
Set the collected information stored in the Context as `metadata` of a `Logger`. This `Logger` will be exposed as part of the `Context` API. *When logging messages using said `Logger` the stored metadata will then be logged too.*

slashmo:
This is a draft implementation of a `BaggageContext` `logger` property that is automatically populated with the metadata stored in the context.

Eventually closes #7

### DIFF
--- a/Sources/Baggage/BaggageContext.swift
+++ b/Sources/Baggage/BaggageContext.swift
@@ -41,11 +41,21 @@ public struct BaggageContext {
         get {
             self._storage[ObjectIdentifier(key)]?.forceUnwrap(key)
         } set {
-            self._storage[ObjectIdentifier(key)] = newValue == nil ? nil : ValueContainer(value: newValue!)
+            self._storage[ObjectIdentifier(key)] = newValue == nil ? nil : ValueContainer(keyName: String(describing: key.self), value: newValue!)
         }
     }
 
+    public var printableMetadata: [String: CustomStringConvertible] {
+        var description = [String: CustomStringConvertible]()
+        for (_, value) in self._storage {
+            // TODO: key could be not unique
+            description[value.keyName] = String(describing: value.value)
+        }
+        return description
+    }
+
     private struct ValueContainer {
+        let keyName: String
         let value: Any
 
         func forceUnwrap<Key: BaggageContextKey>(_ key: Key.Type) -> Key.Value {
--- a/Sources/Baggage/BaggageContext.swift
+++ b/Sources/Baggage/BaggageContext.swift
@@ -42,23 +42,20 @@ public struct BaggageContext {
             self._storage[AnyBaggageContextKey(key)]?.forceUnwrap(key)
         } set {
             self._storage[AnyBaggageContextKey(key)] = newValue.map {
-                ValueContainer(value: $0, key: AnyBaggageContextKey(key))
+                ValueContainer(value: $0)
             }
         }
     }
 
     public var baggageItems: [AnyBaggageContextKey: Any] {
-        var description = [AnyBaggageContextKey: Any]()
-        for (_, value) in self._storage {
-            // TODO: key may not be unique
-            description[value.key] = value.value
+        // TODO: key may not be unique
+        self._storage.reduce(into: [:]) {
+            $0[$1.key] = $1.value.value
         }
-        return description
     }
 
     private struct ValueContainer {
         let value: Any
-        let key: AnyBaggageContextKey
 
         func forceUnwrap<Key: BaggageContextKey>(_ key: Key.Type) -> Key.Value {
             self.value as! Key.Value
--- a/Sources/Baggage/BaggageContext.swift
+++ b/Sources/Baggage/BaggageContext.swift
@@ -89,7 +89,7 @@ public struct AnyBaggageContextKey {
     }
 
     public init<Key>(_ keyType: Key.Type) where Key: BaggageContextKey {
-        self.keyType = keyType.self
+        self.keyType = keyType
         self._name = keyType.name
     }
 }
--- a/Sources/Baggage/BaggageContext.swift
+++ b/Sources/Baggage/BaggageContext.swift
@@ -41,7 +41,10 @@ public struct BaggageContext {
         get {
             self._storage[ObjectIdentifier(key)]?.forceUnwrap(key)
         } set {
-            self._storage[ObjectIdentifier(key)] = newValue == nil ? nil : ValueContainer(keyName: String(describing: key.self), value: newValue!)
+            self._storage[ObjectIdentifier(key)] = newValue.map {
+                // TODO: consider if keys should allow optional string names; static property on the type?
+                ValueContainer(keyName: String(describing: key.self), value: $0)
+            }
         }
     }
 
--- a/Sources/Baggage/BaggageContext.swift
+++ b/Sources/Baggage/BaggageContext.swift
@@ -32,34 +32,33 @@
 ///       }
 ///     }
 public struct BaggageContext {
-    private var _storage = [ObjectIdentifier: ValueContainer]()
+    private var _storage = [AnyBaggageContextKey: ValueContainer]()
 
     /// Create an empty `BaggageContext`.
     public init() {}
 
     public subscript<Key: BaggageContextKey>(_ key: Key.Type) -> Key.Value? {
         get {
-            self._storage[ObjectIdentifier(key)]?.forceUnwrap(key)
+            self._storage[AnyBaggageContextKey(key)]?.forceUnwrap(key)
         } set {
-            self._storage[ObjectIdentifier(key)] = newValue.map {
-                // TODO: consider if keys should allow optional string names; static property on the type?
-                ValueContainer(keyName: String(describing: key.self), value: $0)
+            self._storage[AnyBaggageContextKey(key)] = newValue.map {
+                ValueContainer(value: $0, key: AnyBaggageContextKey(key))
             }
         }
     }
 
-    public var printableMetadata: [String: CustomStringConvertible] {
-        var description = [String: CustomStringConvertible]()
+    public var baggageItems: [AnyBaggageContextKey: Any] {
+        var description = [AnyBaggageContextKey: Any]()
         for (_, value) in self._storage {
-            // TODO: key could be not unique
-            description[value.keyName] = String(describing: value.value)
+            // TODO: key may not be unique
+            description[value.key] = value.value
         }
         return description
     }
 
     private struct ValueContainer {
-        let keyName: String
         let value: Any
+        let key: AnyBaggageContextKey
 
         func forceUnwrap<Key: BaggageContextKey>(_ key: Key.Type) -> Key.Value {
             self.value as! Key.Value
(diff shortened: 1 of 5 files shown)


Sources/Baggage/BaggageContext.swift
```diff
@@ -41,11 +41,21 @@ public struct BaggageContext {
         get {
             self._storage[ObjectIdentifier(key)]?.forceUnwrap(key)
         } set {
-            self._storage[ObjectIdentifier(key)] = newValue == nil ? nil : ValueContainer(value: newValue!)
+            self._storage[ObjectIdentifier(key)] = newValue == nil ? nil : ValueContainer(keyName: String(describing: key.self), value: newValue!)
```
ktoso: `!`-alert! ⚠️ 

You can avoid it by doing:

```swift
            self._storage[ObjectIdentifier(key)] = newValue.map {
                ValueContainer(keyName: String(describing: key.self), value: $0)
            }
```

I would also propose:

```suggestion
            self._storage[ObjectIdentifier(key)] = newValue.map {
                // TODO: consider if keys should allow optional string names; static property on the type?
                ValueContainer(keyName: String(describing: key.self), value: $0)
            }
```

since we may consider that

slashmo: Lol, I use your first suggestion all the time, but somehow forgot to use it here. Thanks! 👍

Sources/BaggageLogging/Baggage+Logger.swift
```diff
@@ -0,0 +1,12 @@
+import Baggage
+import Logging
+
+extension BaggageContext {
+    public var logger: Logger {
+        var logger = Logger(label: "BaggageContext")
```
ktoso: I "fear" we may need to merry a few concepts here, specifically:

- some applications like to "create root logger and always pass it around and never use the global factory, and
- some applications don't care, and they'd be ok with us making up a logger like that.

What if we indeed try to allow the 1st case as well... Granted this is somewhat exploratory territory but let's give it a shot. 

It would indeed start moving us towards "context is a bag of random stuff" which I'm worried about but the upside is better composition and this pattern as well...

Say, what if we allow:

```suggestion
        var logger = self[BaseLoggerKey.self] ?? Logger(label: "BaggageContext")
```

this way a framework could still adopt this pattern...

Not sure if either style will be "the style" that will be adopted in the end, but we should explore these.

Let's also here on the documentation explain how it would be used and what it does.

With regards to copying all the metadata into the logger eagerly... I think that's where a custom logging handler could play a role; there is the `    internal init(label: String, _ handler: LogHandler)` overload which would allow us to implement the "materialize (make a string from it) values from context _only_ when we actually log things". Worth exploring!

slashmo: > It would indeed start moving us towards "context is a bag of random stuff"

This is really interesting. I didn't think about using the fact that we have "storage" at hand to "inject" a logger, pretty smart. However, as you already mentioned, we are somewhat crossing a line with this. But let's explore this further 👍

> I think that's where a custom logging handler could play a role

Interesting as well. I'll look into this too.

Sources/Baggage/BaggageContext.swift
```diff
@@ -41,11 +41,21 @@ public struct BaggageContext {
...
     }
 
+    public var printableMetadata: [String: CustomStringConvertible] {
+        var description = [String: CustomStringConvertible]()
+        for (_, value) in self._storage {
+            // TODO: key could be not unique
+            description[value.keyName] = String(describing: value.value)
+        }
+        return description
+    }
```
ktoso, line 55: That's too specific to our specific "we need to print those" need IMHO.

It also means that we have:

- to allocate a dictionary which we may not want to do (what if we only are looking for 1 key, no need to look for the rest if we found it)
- we actually render the values, even if we'd never print them -- rendering (`string(describing:`) for them might be expensive (say, storage is integers, but printing them requires base64ing them, would be a waste to make this on every log statement even if we never printed them etc.

What if we get inspired by OpenTracing here: https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/SpanContext.java#L59

```
    Iterable<Map.Entry<String, String>> baggageItems();
```

so return an iterable with the rendered key and `Any` perhaps in our case?

Sources/Baggage/BaggageContext.swift
```diff
@@ -41,11 +41,24 @@ public struct BaggageContext {
         get {
             self._storage[ObjectIdentifier(key)]?.forceUnwrap(key)
         } set {
-            self._storage[ObjectIdentifier(key)] = newValue == nil ? nil : ValueContainer(value: newValue!)
+            self._storage[ObjectIdentifier(key)] = newValue.map {
+                // TODO: consider if keys should allow optional string names; static property on the type?
+                ValueContainer(keyName: String(describing: key.self), value: $0)
```
ktoso: iffy about the `ValueContainer` _having_ the name -- that means that:
- every time we _set_ values we need to materialize the string description of the type -- that may be expensive. Again, let's only do it when we can i.e. we'll need to store the key type...

We can do this if we use a special:

```swift
struct AnyKey: Hashable { 
  let keyType: Any.Type
  var id: ObjectIdentifier { ObjectIdentifier(self.keyType) } // this is cheap
  let name: String 
  init<Key>(_ key: Key) where Key: BaggageContextKey { 
    // we can do the if the Key has a static name we could store it; otherwise we can do the description() call here
    ...
   }

  // (hash and equals by the Objectidentifier _only_)
}
```

as key in our `_storage` this way we can use the Any.Type to get its name when we need to (string describing).
We could also experiment with static properties for the names if we needed to...?

Related notes:

- "Strings under 15 utf8 bytes doesn't heap allocate (on 64 bit)" 
  - good for us, most ids should be pretty short i hope
- `Objectidentifier(type)` is cheap and boils down to a (supported) unsafeBitCast

Sources/BaggageLogging/Baggage+Logger.swift
```diff
@@ -3,10 +3,17 @@ import Logging
 
 extension BaggageContext {
     public var logger: Logger {
-        var logger = Logger(label: "BaggageContext")
-        for (key, value) in printableMetadata {
+        var logger = self[BaseLoggerKey.self] ?? Logger(label: "BaggageContext")
+        // TODO: Filtering out a specific key to get rid of the logger seems very strange
```
ktoso, line 7: agreed hmmm, good todo, lets think about it

Sources/BaggageLogging/Baggage+Logger.swift
```diff
@@ -3,10 +3,17 @@ import Logging
 
 extension BaggageContext {
     public var logger: Logger {
-        var logger = Logger(label: "BaggageContext")
-        for (key, value) in printableMetadata {
+        var logger = self[BaseLoggerKey.self] ?? Logger(label: "BaggageContext")
```
ktoso, line 6: cool; let's see if this pattern is useful or not 👍

Sources/Baggage/BaggageContext.swift
```diff
@@ -55,7 +64,42 @@ public struct BaggageContext {
 }
 
 /// `BaggageContextKey`s are used as keys in a `BaggageContext`. Their associated type `Value` gurantees type-safety.
+/// To give your `BaggageContextKey` an explicit name you may override the `name` property.
 public protocol BaggageContextKey {
     /// The type of `Value` uniquely identified by this key.
     associatedtype Value
+
+    /// The human-readable name of this key. Defaults to `nil`.
+    static var name: String? { get }
```
ktoso, line 73: Excellent 👍

Sources/Baggage/BaggageContext.swift
```diff
@@ -55,7 +64,42 @@ public struct BaggageContext {
...
+}
+
+extension AnyBaggageContextKey: Hashable {
+    public static func == (lhs: AnyBaggageContextKey, rhs: AnyBaggageContextKey) -> Bool {
+        ObjectIdentifier(lhs.keyType) == ObjectIdentifier(rhs.keyType)
+    }
+
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(ObjectIdentifier(self.keyType))
+    }
```
ktoso, line 104: 👍

Sources/Baggage/BaggageContext.swift
```diff
@@ -32,16 +32,25 @@
...
         } set {
-            self._storage[ObjectIdentifier(key)] = newValue == nil ? nil : ValueContainer(value: newValue!)
+            self._storage[AnyBaggageContextKey(key)] = newValue.map {
+                ValueContainer(value: $0)
+            }
+        }
+    }
+
+    public var baggageItems: [AnyBaggageContextKey: Any] {
+        // TODO: key may not be unique
```
ktoso, line 51: does this TODO still hold?

Sources/Baggage/BaggageContext.swift
```diff
@@ -32,16 +32,25 @@
...
+        self._storage.reduce(into: [:]) {
+            $0[$1.key] = $1.value.value
         }
```
ktoso, line 54: I'd kind of hope we would not need to allocate a new dictionary, but we can solve this in a separate PR; we can also add some performance benchmarks and then solve that; ticket perhaps for this?

Sources/Baggage/BaggageContext.swift
```diff
@@ -55,7 +64,42 @@ public struct BaggageContext {
...
+    private let _name: String?
+
+    /// A human-readable String representation of the underlying key.
+    /// If no explicit name has been set on the wrapped key the type name is used.
+    public var name: String {
+        self._name ?? String(describing: self.keyType.self)
+    }
+
+    public init<Key>(_ keyType: Key.Type) where Key: BaggageContextKey {
+        self.keyType = keyType.self
```
ktoso:
```suggestion
        self.keyType = keyType
```

is enough AFAIK?

Sources/BaggageLogging/Baggage+Logger.swift
```diff
@@ -0,0 +1,19 @@
+import Baggage
+import Logging
+
+extension BaggageContext {
+    public var logger: Logger {
+        var logger = self[BaseLoggerKey.self] ?? Logger(label: "BaggageContext")
+        // TODO: Filtering out a specific key to get rid of the logger seems very strange
+        for (key, value) in baggageItems where key.keyType != BaseLoggerKey.self {
```
ktoso, line 8: ticket or some writeup about this problem somewhere?

What if we go down the route of "context is a bag of ANYTHING" then there could be random stuff in here including functions etc which we'd also not want to make into a metadata. Would we need to make some other type of key for "Actual metadata" and "terrible closure" and such things would have to be aware of it?

That's quite a problem if we allow random stuff in contexts other than "just data"... What would be another way we can solve this?

Sources/BaggageLogging/Baggage+Logger.swift
```diff
@@ -0,0 +1,19 @@
...
+            logger[metadataKey: key.name] = "\(String(describing: value))"
+        }
```
ktoso, line 10: Any ideas how on the LogHandler backend that could delegate "to" context so we would not have to create these rendered values each time, but only when the logger actually logs.

Say:

```swift
// assume `context` has a lot of values stored

context.log.trace("is not going to be logged") // ⚠️ oh oh!
// but with current impl we allocated all the string reprs for all the strings!
```

Is there some way we can avoid this with having a LogHandler?

ktoso: This can be a separate ticket and PR :)

Sources/Baggage/BaggageContext.swift
slashmo: Yes, will be included in the "tradeoffs" ticket after the merge.

PR #45, opened 2020-06-22
Avoid trouble with shutting down HTTPClient

ktoso:


UseCases/Sources/HTTPEndToEnd/InstrumentedHTTPClient.swift
```diff
@@ -6,21 +6,36 @@ import NIO
 import NIOHTTP1
 
 struct InstrumentedHTTPClient {
-    private let client = HTTPClient(eventLoopGroupProvider: .createNew)
```
ktoso, line 9: we were ignoring the init's passed in value here; so fixed this

UseCases/Sources/HTTPEndToEnd/InstrumentedHTTPClient.swift
```diff
@@ -6,21 +6,36 @@ import NIO
...
 
+    // TODO: deadline: NIODeadline? would move into baggage?
+    public func get(url: String, baggage: BaggageContext = .init()) -> EventLoopFuture<HTTPClient.Response> {
+        do {
+            let request = try HTTPClient.Request(url: url, method: .GET)
+            return self.execute(request: request, baggage: baggage)
+        } catch {
+            return self.client.eventLoopGroup.next().makeFailedFuture(error)
+        }
+    }
```
ktoso, line 29: just so it can be used raw in main.swift

UseCases/Sources/HTTPEndToEnd/InstrumentedHTTPClient.swift
```diff
@@ -6,21 +6,36 @@ import NIO
...
     init<I>(instrument: I, eventLoopGroupProvider: HTTPClient.EventLoopGroupProvider)
         where
         I: InstrumentProtocol,
         I.InjectInto == HTTPHeaders,
         I.ExtractFrom == HTTPHeaders {
+        self.client = HTTPClient(eventLoopGroupProvider: eventLoopGroupProvider)
         self.instrument = Instrument(instrument)
     }
 
+    // TODO: deadline: NIODeadline? would move into baggage?
```
ktoso, line 21: that's a good future question -- maybe ticket?

Maybe we'll make a fork of HTTPClient soon and propose how it'd look like with context?

UseCases/Sources/HTTPEndToEnd/InstrumentedHTTPClient.swift
```diff
@@ -6,21 +6,36 @@ import NIO
...
     func execute(request: HTTPClient.Request, baggage: BaggageContext) -> EventLoopFuture<HTTPClient.Response> {
         var request = request
         self.instrument.inject(from: baggage, into: &request.headers)
         baggage.logger.info("AsyncHTTPClient: Execute request")
         return self.client.execute(request: request)
     }
+
+    func syncShutdown() throws {
+        try self.client.syncShutdown()
+    }
```
ktoso, line 40: this cannot be called on an event loop, which makes it hard it the Handler is the "only" place that is able to call shutdown -- we could not call it in there after all. Thus I propose to pass in the client to handlers

UseCases/Sources/HTTPEndToEnd/Services/StorageServiceHandler.swift
```diff
@@ -9,11 +9,12 @@ final class StorageServiceHandler: ChannelInboundHandler {
     typealias InboundIn = HTTPServerRequestPart
     typealias OutboundOut = HTTPServerResponsePart
 
+    let logger = Logger(label: "StorageService")
```
ktoso, line 12: feel free to ignore this bit... Just habit I guess, previous is ok too

UseCases/Sources/HTTPEndToEnd/main.swift
```diff
@@ -9,15 +9,15 @@ import NIOHTTP1
 import NIO
 
 let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
-let threadPool = NIOThreadPool(numberOfThreads: 6)
-threadPool.start()
```
ktoso, line 13: i don't think we'll need one?

UseCases/Sources/HTTPEndToEnd/main.swift
```diff
@@ -57,8 +56,9 @@ httpClient.get(url: "http://localhost:8080").whenComplete { result in
 sleep(20)
 
 try httpClient.syncShutdown()
+try orderServiceChannel.close().wait()
+try storageServiceChannel.close().wait()
```
ktoso, line 60: I guess we don't really have to right now since we kill the EL, but this would be a good way if we had any cleanup to do in `channelRemoved` -- but we don't... So consider this optional here I guess

UseCases/Sources/HTTPEndToEnd/InstrumentedHTTPClient.swift
slashmo: Wow, how did I miss that? 🤦‍♂️😁

UseCases/Sources/HTTPEndToEnd/main.swift
```diff
@@ -9,15 +9,15 @@ import NIOHTTP1
...
+
+let httpClient = InstrumentedHTTPClient(instrument: FakeTracer(), eventLoopGroupProvider: .createNew)
 
 let orderServiceBootstrap = ServerBootstrap(group: eventLoopGroup)
     .serverChannelOption(ChannelOptions.backlog, value: 256)
     .serverChannelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
     .childChannelInitializer { channel in
         channel.pipeline.configureHTTPServerPipeline().flatMap {
-            channel.pipeline.addHandler(OrderServiceHandler(instrument: FakeTracer()))
+            channel.pipeline.addHandler(OrderServiceHandler(httpClient: httpClient, instrument: FakeTracer()))
```
slashmo, line 20: 👍

slashmo (approved):

UseCases/Sources/HTTPEndToEnd/InstrumentedHTTPClient.swift
ktoso: happens :)

Issue #37, opened 2020-06-18
BaggageContext and Swift Logging: Design alternatives


slashmo:
Logging support has now been implemented as part of a `BaggageLogging` library (that depends on `swift-log`). It allows for [setting a base Logger](https://github.com/slashmo/gsoc-swift-tracing/blob/main/Tests/BaggageLoggingTests/BaggageLoggingTests.swift#L37) that is then being used to compute the `context.logger` property. In the current implementation, this base `Logger` is injected into the `BaggageContext` by stuffing it in the baggage via a special `BaggageContext.BaseLoggerKey`. 

Below are some pro/cons on this approach and a possible alternative.

#### Pros 👍

- It's easy, no need to pass a `Logger` while initializing a `BaggageContext`
- It allows us to add the `logger` property in a separate library, as the `BaggageContext` itself becomes the "property storage"

#### Cons 👎

- When setting the metadata on the `Logger` [we had to manually exclude](https://github.com/slashmo/gsoc-swift-tracing/blob/main/Sources/BaggageLogging/Baggage%2BLogger.swift#L20) the `BaggageContext.BaseLoggerKey` so that it's not included in the log
- This may set an example for others to stuff in non-metadata related things into the `BaggageContext`
- It feels like a hack because we weren't able to add the `Logger` as a stored property in the `BaggageContext` extension

#### Pros 👍

- Feels more natural, no need to make an exception for the otherwise metadata-only `BaggageContext`
- Doesn't lead to exclusion of certain keys when adding to the `Logger` metadata

#### Cons 👎

- `BaggageContext` would need to depend on `swift-log` directly

---

What are your thoughts on this. Other pros/cons, or even other alternatives to this approach?

ktoso:
More approaches to consider (and I guess we may need to PoC around).

(I'm not really proposing the below, does not feel right, but let's list potential solutions before we decide on one)

### (3) Logger to be _created_ or "updated with" a Context

This would take the form of something like

```swift
func doThings(..., context: BaggageContext) { 
    Logger("label", context).info("hello") // ad-hoc creating, some projects do not want that...
}

func doThings(..., log: Logger, context: BaggageContext) { 
    logger(context).info("hello") // "app wide passed around logger" and "enrich" with context
} // this can get confusing perhaps?
```

Would it be possible perhaps to avoid this with the below?

```swift
func doThings(..., logger: BaggageLogger) { 
  logger.baggage // ...
  logger.info("xxx") // ...
[...]

PR #49, opened 2020-06-23
Print keys in custom BaggageContext description

### LINKED ISSUE #41: Consider nicer string description of baggage context
Current printout just uses the default properties which is `BaggageContext(_storage: [:])` which reads a bit meh, since we want to hide that _storage.

Specifically, it looks like this today:

`BaggageContext(_storage: [Baggage.AnyBaggageContextKey(keyType: MyModule.MyFunKey, _name: nil): Baggage.BaggageContext.(unknown context at $10773422c).ValueContainer(value: ...)])`

That's a lot of noise :) 

Perhaps the default rendering should only render the keys?

slashmo:
Closes #41 

With this in place, `BaggageLogging`s "hack" of storing the `Logger` in the `BaggageContext` feels even weirder, because the `BaseLoggerKey` is clearly readable in the output 🤔 (#37)

### DIFF
--- a/Sources/Baggage/BaggageContext.swift
+++ b/Sources/Baggage/BaggageContext.swift
@@ -63,6 +63,12 @@ public struct BaggageContext {
     }
 }
 
+extension BaggageContext: CustomStringConvertible {
+    public var description: String {
+        "\(Self.self)(keys: \(self._storage.map(\.key.name)))"
+    }
+}
+
 /// `BaggageContextKey`s are used as keys in a `BaggageContext`. Their associated type `Value` gurantees type-safety.
 /// To give your `BaggageContextKey` an explicit name you may override the `name` property.
 public protocol BaggageContextKey {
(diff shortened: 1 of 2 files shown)


slashmo:
@ktoso What's your opinion on unit-testing things like this? I left it out because this is basically just a debugging tool, but am happy to add a test covering this as well.

ktoso:
> With this in place, BaggageLoggings "hack" of storing the Logger in the BaggageContext feels even weirder, because the BaseLoggerKey is clearly readable in the output 🤔 (#37)

Yeah, we'll need to rethink this it definitely is too weird to be _the way_ I'm begining to feel...

> @ktoso What's your opinion on unit-testing things like this? I left it out because this is basically just a debugging tool, but am happy to add a test covering this as well.

I would actually test such a thing. I tend to think about "the command line output" as the "UI" of libraries (or DX as some call it, Developer Experience (like UX)), so if someone is using the lib, and we want to make sure they have a good time even in such "side" tasks I think it's worth testing it. I even write tests for "bla bla action causes a nice and understandable log message" and similar things sometimes :) 

+1 for a simple test.

ktoso (commented):
Adding a mini test would be nice :)

ktoso (approved):
LGTM :)

PR #53, opened 2020-06-26
Rename Instrument Box type to AnyInstrument

### LINKED ISSUE #43: Instrument "box" naming
(ignoring how the inject/extract to/from are expressed).

We'd want to write Instrument as the generic rather than `I` since that's common swift style to use full words.
But we today use Instrument for the wrapper type:

```swift
    init<Instrument>(instrument: Instrument)
        where Instrument: InstrumentProtocol,
        Instrument.InjectInto == HTTPHeaders,
        Instrument.ExtractFrom == HTTPHeaders {

        self.instrument = Instrument(instrument) // conflict
```

What would be a better naming strategy for this type:

```swift
/// A box-type for an `InstrumentProtocol`, necessary for creating homogeneous collections of `InstrumentProtocol`s.
public struct Instrument<InjectInto, ExtractFrom>: InstrumentProtocol {
    private let inject: (BaggageContext, inout InjectInto) -> Void
    private let extract: (ExtractFrom, inout BaggageContext) -> Void

    /// Wrap the given `InstrumentProtocol` inside an `Instrument`.
    /// - Parameter instrument: The `InstrumentProtocol` being wrapped.
    public init<I>(_ instrument: I)
        where
        I: InstrumentProtocol,
        I.InjectInto == InjectInto,
        I.ExtractFrom == ExtractFrom {
```

`TracingInstrument` does come to mind, but these ones are _not_ specifically tracing, they are generic.

Having looked at this again I guess this should be `AnyInstrument` and the protocol can become `InstrumentProtocol` (OR `BaggageInstrument`)... WDYT?

slashmo:
Closes #43

### DIFF
--- a/Sources/Instrumentation/Instrument.swift
+++ b/Sources/Instrumentation/Instrument.swift
@@ -25,17 +25,17 @@ public protocol InstrumentProtocol {
 }
 
 /// A box-type for an `InstrumentProtocol`, necessary for creating homogeneous collections of `InstrumentProtocol`s.
-public struct Instrument<InjectInto, ExtractFrom>: InstrumentProtocol {
+public struct AnyInstrument<InjectInto, ExtractFrom>: InstrumentProtocol {
     private let inject: (BaggageContext, inout InjectInto) -> Void
     private let extract: (ExtractFrom, inout BaggageContext) -> Void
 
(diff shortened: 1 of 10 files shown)


ktoso (approved):
👍 Looking good :)

PR #55, opened 2020-06-30
Introduce Carrier, Injector, and Extractor types

### LINKED ISSUE #27: Consider renaming ExtractFrom/InjectInto to Source/Destination
Via @tanner0101

> The associated types InjectInto and ExtractFrom are prepositions, not nouns. Maybe something like Source / Destination would be better?

### LINKED ISSUE #54: Extract "InjectInto/ExtractFrom" from InstrumentProtocol
As mentioned in #47

> While working on this I noticed that the current shape of `InstrumentProtocol` with its associated types `InjectInto` & `ExtractFrom` prohibits us from setting up a singleton instance to be bootstrapped, as we can't know upfront what the user will bootstrap the system with (so what's the properties type?). 

> Also, it's definitely possible for a system to use different technologies (HTTP, gRPC, ...) being instrumented by the same cross-cutting tool. That's why it probably makes sense to not have the `Instrument` be generic over the `InjectInto`/`ExtractFrom` types, but rather to move this responsibility somewhere else.

slashmo:
- Closes #54
- Closes #27 (through removing them 😊)


Sources/NIOInstrumentation/BaggageContextInboundHTTPHandler.swift
```diff
@@ -7,22 +7,18 @@ public final class BaggageContextInboundHTTPHandler: ChannelInboundHandler {
...
+    public init(instrument: Instrument, onBaggage: @escaping (BaggageContext) -> Void) {
+        self.instrument = instrument
         self.onBaggageExtracted = onBaggage
     }
 
     public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
         guard case .head(let head) = unwrapInboundIn(data) else { return }
         var baggage = BaggageContext()
-        self.instrument.extract(from: head.headers, into: &baggage)
+        self.instrument.extract(head.headers, into: &baggage, using: HTTPHeadersExtractor())
```
ktoso, line 21: excellent 👍

Sources/Instrumentation/MultiplexInstrument.swift
```diff
@@ -1,25 +1,35 @@
 import Baggage
 
-/// A pseudo-`Instrument` that may be used to instrument using multiple other `AnyInstrument`s
-/// across a common `BaggageContext`.
-public struct MultiplexInstrument<InjectInto, ExtractFrom> {
-    private var instruments: [AnyInstrument<InjectInto, ExtractFrom>]
+/// A pseudo-`Instrument` that may be used to instrument using multiple other `Instrument`s across a
+/// common `BaggageContext`.
+public struct MultiplexInstrument {
+    private var instruments: [Instrument]
```
ktoso, line 6: great, no longer fixed to specific Into/Froms.

Sources/Instrumentation/Instrument.swift
```diff
@@ -1,50 +1,58 @@
...
 
-/// Conforming types are usually cross-cutting tools like tracers that extract values of one type (`ExtractFrom`) into a
-/// `BaggageContext` and inject values stored in a `BaggageContext` into another type (`InjectInto`).
-/// `ExtractFrom` and `InjectInto` may very well be of the same type, e.g. when a cross-cutting tool propagates
-/// baggage through HTTP headers.
-public protocol InstrumentProtocol {
-    associatedtype InjectInto
-    associatedtype ExtractFrom
+/// Conforming types are used to carry metadata over a boundary (e.g. HTTP).
+public protocol CarrierProtocol {}
```
ktoso: I'm not sure we need this and it may have negative compatibility implications ... 
I need to sanity check if I remember this right though, will loop back soon

Sources/NIOInstrumentation/HTTPHeadersCarrier.swift
```diff
@@ -0,0 +1,20 @@
...
+    }
+}
+
+public final class HTTPHeadersInjector: InjectorProtocol {
+    public init() {}
+
+    public func inject(_ value: String, forKey key: String, into headers: inout HTTPHeaders) {
+        headers.replaceOrAdd(name: key, value: value)
+    }
+}
```
ktoso, line 18: 👍

Sources/NIOInstrumentation/HTTPHeadersCarrier.swift
```diff
@@ -0,0 +1,20 @@
...
+public final class HTTPHeadersExtractor: ExtractorProtocol {
+    public init() {}
+
+    public func extract(key: String, from headers: HTTPHeaders) -> String? {
+        headers.first(name: key)
+    }
+}
+
+public final class HTTPHeadersInjector: InjectorProtocol {
+    public init() {}
```
ktoso, line 13: can those be structs? :)

Sources/NIOInstrumentation/HTTPHeadersCarrier.swift
```diff
@@ -0,0 +1,20 @@
+import Instrumentation
+import NIOHTTP1
+
+extension HTTPHeaders: CarrierProtocol {}
+
+public final class HTTPHeadersExtractor: ExtractorProtocol {
```
ktoso: can those be structs? :)

Sources/Instrumentation/MultiplexInstrument.swift
slashmo: Yep, way nicer to work with now 👍

Sources/Instrumentation/Instrument.swift
ktoso: Ah okey the thing I remembered is "just" https://github.com/apple/swift-nio/blob/master/docs/public-api.md#2-conforming-nio-types-to-protocols 

which is a bit of a mix for us and mostly okey... since "we" the tracing impls providing the conformance would do this.

Having that said, I don't think we need this protocol -- we can express everything with "just" `associatedtype Carrier` without the `: CarrierProtocol`. let's remove to keep the API minimal?

ktoso (approved):
Looks good! I think we can do without the `CarrierProtocol`, otherwise LGTM :)

Sources/Instrumentation/Instrument.swift
slashmo: Yes, let's remove it 👌

PR #56, opened 2020-06-30
Add InstrumentationSystem

### LINKED ISSUE #47: Consider InstrumentationSystem (?) similar to MetricsSystem et al
Libraries will often need to pull an instrument "out of thin air" in a similar way as metrics have to be pulled out sometimes.

We should offer something similar to https://github.com/apple/swift-metrics#selecting-a-metrics-backend-implementation-applications-only

Naming will evolve here but something like `InstrumentationSystem` to be consistent with Metrics and logging perhaps?
We should also think how it relates to tracing; would we allow e.g. `InstrumentationSystem.instrument` and also `InstrumentationSystem.tracer` (where a tracer is-a instrument but not all usages need a tracer etc.)

slashmo:
Closes #47

### DIFF
--- a/Sources/Instrumentation/InstrumentationSystem.swift
+++ b/Sources/Instrumentation/InstrumentationSystem.swift
@@ -1,0 +1,47 @@
+import Baggage
+
+/// `InstrumentationSystem` is a global facility where the default cross-cutting tool can be configured.
+/// It is set up just once in a given program select the desired `Instrument` implementation.
+///
+/// - Note: If you need to use more that one cross-cutting tool you can do so by using `MultiplexInstrument`.
+public enum InstrumentationSystem {
+    private static let lock = ReadWriteLock()
+    private static var _instrument: Instrument = NOOPInstrument.instance
+    private static var isInitialized = false
+
+    /// Globally select the desired `Instrument` implementation.
+    ///
+    /// - Parameter instrument: The `Instrument` you want to share globally within your system.
+    /// - Warning: Do not call this method more than once. This will lead to a crash.
+    public static func bootstrap(_ instrument: Instrument) {
+        self.lock.withWriterLock {
+            precondition(
+                !self.isInitialized, """
+                InstrumentationSystem can only be initialized once per process. Consider using MultiplexInstrument if
+                you need to use multiple instruments.
+                """
+            )
+            self._instrument = instrument
+            self.isInitialized = true
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


slashmo:
@ktoso  I copied over the `Lock` parts we needed from Swift Metrics. Should we start adding license headers now?

ktoso:
> @ktoso I copied over the Lock parts we needed from Swift Metrics.

ok good

> Should we start adding license headers now?

Yes, about time I guess.

You could copy the `validate_license_header.sh` script as well and adjust it to our project name from metrics.

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -0,0 +1,47 @@
...
+    ///
+    /// - Parameter instrument: The `Instrument` you want to share globally within your system.
+    /// - Warning: Do not call this method more than once. This will lead to a crash.
+    public static func bootstrap(_ instrument: Instrument) {
+        self.lock.withWriterLock {
+            precondition(
+                !self.isInitialized, """
+                InstrumentationSystem can only be initialized once per process. Consider using MultiplexInstrument if
+                you need to use multiple instruments.
+                """
```
ktoso, line 22: 👍

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    /// Returns the globally configured `Instrument`. Defaults to a no-op `Instrument` if `boostrap` wasn't called before.
+    public static var instrument: Instrument {
+        self.lock.withReaderLock { self._instrument }
+    }
+}
+
+private final class NOOPInstrument: Instrument {
```
ktoso, line 35: 👍 maybe in its own file though?

Sources/Instrumentation/Locks.swift
```diff
@@ -0,0 +1,89 @@
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
```
ktoso, line 1: copy license header as well from where these files are taken? (metrics or logging)

ktoso (approved):
LGTM :) Maybe split out the NOOP impl into its own file.

Follow up with license headers?

Sources/Instrumentation/InstrumentationSystem.swift
slashmo: Then we'd have to make it internal. Would that be worth it?

slashmo: Are there other uses of such a `NOOPInstrument`?

PR #62, opened 2020-07-07
[WIP] Add TracingInstrument

### LINKED ISSUE #50: Design initial TracingInstrument protocol
Inherits from `Instrument` to require span-related APIs

### LINKED ISSUE #68: SpanAttributes as a type
Since the spec suggests those should keep order, just a plain old dict is not going serve this well:

> Attributes SHOULD preserve the order in which they're set. Setting an attribute with the same key as an existing attribute SHOULD overwrite the existing attribute's value.

This basically means a SortedDict which we don't have;

It's a SHOULD so we can move forward without it for now but we want to be in a place where we can swap the impl without breaking anyone to conform to this requirement.

So, let's:

```swift
struct SpanAttributes { 
    subscript(_ name: String) -> SpanAttribute? 
    func forEachAttribute(_ callback: (String, SpanAttribute) -> ())
}
```

and for now back it with a `private [String:SpanAttribute]`  which we can easily replace eventually

slashmo:
This is a super early WIP for #50 that should help us start discussing more specific design decisions.

- Closes #50
- Closes #68

## Todos

- [x] Create `SpanAttributes` wrapper struct
- [x] Add `Span` Status API: [OT Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-status)
- [x] Add `Span.isRecording` API: [OT Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#isrecording)
- [x] Write documentation for `TracingInstrument` and related types

### DIFF
--- a/Sources/Instrumentation/Instrument.swift
+++ b/Sources/Instrumentation/Instrument.swift
@@ -68,7 +68,6 @@ public protocol Instrument {
 }
 
 public struct NoOpInstrument: Instrument {
-
     public func inject<Carrier, Injector>(_ baggage: BaggageContext, into carrier: inout Carrier, using injector: Injector)
         where
         Injector: InjectorProtocol,
--- a/Sources/Instrumentation/Instrument.swift
+++ b/Sources/Instrumentation/Instrument.swift
@@ -66,3 +66,16 @@ public protocol Instrument {
         Injector: InjectorProtocol,
         Injector.Carrier == Carrier
 }
+
+public struct NoOpInstrument: Instrument {
+
+    public func inject<Carrier, Injector>(_ baggage: BaggageContext, into carrier: inout Carrier, using injector: Injector)
+        where
+        Injector: InjectorProtocol,
+        Carrier == Injector.Carrier {}
+
+    public func extract<Carrier, Extractor>(_ carrier: Carrier, into baggage: inout BaggageContext, using extractor: Extractor)
+        where
+        Extractor: ExtractorProtocol,
+        Carrier == Extractor.Carrier {}
+}
--- a/Sources/Instrumentation/Tracing/NoOpTracing.swift
+++ b/Sources/Instrumentation/Tracing/NoOpTracing.swift
@@ -34,6 +34,7 @@ public struct NoOpTracingInstrument: TracingInstrument {
 
     public struct NoOpSpan: Span {
         public var operationName: String = ""
+        public var status: SpanStatus?
 
         public var startTimestamp: DispatchTime {
             .now()
--- a/Sources/Instrumentation/Tracing/NoOpTracing.swift
+++ b/Sources/Instrumentation/Tracing/NoOpTracing.swift
@@ -14,6 +14,7 @@
 import Baggage
 import Dispatch
 
+/// No operation TracingInstrument, used when no tracing is required.
 public struct NoOpTracingInstrument: TracingInstrument {
     public var currentSpan: Span? = nil
 
--- a/Sources/Instrumentation/Tracing/NoOpTracing.swift
+++ b/Sources/Instrumentation/Tracing/NoOpTracing.swift
@@ -52,15 +52,10 @@ public struct NoOpTracingInstrument: TracingInstrument {
 
         public mutating func addEvent(_ event: SpanEvent) {}
 
-        public var attributes: [String: SpanAttribute] {
-            [:]
-        }
-
-        public subscript(attributeName attributeName: String) -> SpanAttribute? {
+        public var attributes: SpanAttributes {
             get {
-                nil
-            }
-            set {
+                [:]
+            } set {
                 // ignore
             }
         }
--- a/Sources/Instrumentation/Tracing/NoOpTracing.swift
+++ b/Sources/Instrumentation/Tracing/NoOpTracing.swift
@@ -1,0 +1,77 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Tracing open source project
+//
+// Copyright (c) 2020 Moritz Lang and the Swift Tracing project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Baggage
+import Dispatch
+
+public struct NoOpTracingInstrument: TracingInstrument {
+    public var currentSpan: Span? = nil
+
+    public func startSpan(named operationName: String, context: BaggageContext, at timestamp: DispatchTime?) -> Span {
+        NoOpSpan()
+    }
+
+    public func inject<Carrier, Injector>(_ context: BaggageContext, into carrier: inout Carrier, using injector: Injector)
+        where
+        Injector: InjectorProtocol,
+        Carrier == Injector.Carrier {
+    }
+
+    public func extract<Carrier, Extractor>(_ carrier: Carrier, into baggage: inout BaggageContext, using extractor: Extractor)
+        where
+        Extractor: ExtractorProtocol,
+        Carrier == Extractor.Carrier {
+    }
+
+    public struct NoOpSpan: Span {
+        public var operationName: String = ""
+
+        public var startTimestamp: DispatchTime {
+            .now()
+        }
+        public var endTimestamp: DispatchTime? = nil
+
+        public var baggage: BaggageContext {
+            .init()
+        }
+
+        public var events: [SpanEvent] {
+            []
+        }
+
+        public mutating func addEvent(_ event: SpanEvent) {
+        }
+
+        public var attributes: [String: SpanAttribute] {
+            [:]
+        }
+        public subscript(attributeName attributeName: String) -> SpanAttribute? {
+            get {
+                nil
+            }
+            set {
+                // ignore
+            }
+        }
+
+        public var onEnd: (Span) -> Void {
+            { (span) in
+                ()
+            }
+        }
+
+        public mutating func end(at timestamp: DispatchTime) {
+            // ignore
+        }
+    }
+}
--- a/Sources/Instrumentation/Tracing/NoOpTracing.swift
+++ b/Sources/Instrumentation/Tracing/NoOpTracing.swift
@@ -64,11 +64,7 @@ public struct NoOpTracingInstrument: TracingInstrument {
             }
         }
 
-        public var onEnd: (Span) -> Void {
-            { _ in
-                ()
-            }
-        }
+        public let isRecording = false
 
         public mutating func end(at timestamp: DispatchTime) {
             // ignore
--- a/Sources/Instrumentation/Tracing/NoOpTracing.swift
+++ b/Sources/Instrumentation/Tracing/NoOpTracing.swift
@@ -55,7 +55,8 @@ public struct NoOpTracingInstrument: TracingInstrument {
         public var attributes: SpanAttributes {
             get {
                 [:]
-            } set {
+            } 
+            set {
                 // ignore
             }
         }
--- a/Sources/Instrumentation/Tracing/NoOpTracing.swift
+++ b/Sources/Instrumentation/Tracing/NoOpTracing.swift
@@ -24,14 +24,12 @@ public struct NoOpTracingInstrument: TracingInstrument {
     public func inject<Carrier, Injector>(_ context: BaggageContext, into carrier: inout Carrier, using injector: Injector)
         where
         Injector: InjectorProtocol,
-        Carrier == Injector.Carrier {
-    }
+        Carrier == Injector.Carrier {}
 
     public func extract<Carrier, Extractor>(_ carrier: Carrier, into baggage: inout BaggageContext, using extractor: Extractor)
         where
         Extractor: ExtractorProtocol,
-        Carrier == Extractor.Carrier {
-    }
+        Carrier == Extractor.Carrier {}
 
     public struct NoOpSpan: Span {
         public var operationName: String = ""
@@ -39,6 +37,7 @@ public struct NoOpTracingInstrument: TracingInstrument {
         public var startTimestamp: DispatchTime {
             .now()
         }
+
         public var endTimestamp: DispatchTime? = nil
 
         public var baggage: BaggageContext {
@@ -49,12 +48,12 @@ public struct NoOpTracingInstrument: TracingInstrument {
             []
         }
 
-        public mutating func addEvent(_ event: SpanEvent) {
-        }
+        public mutating func addEvent(_ event: SpanEvent) {}
 
         public var attributes: [String: SpanAttribute] {
             [:]
         }
+
         public subscript(attributeName attributeName: String) -> SpanAttribute? {
             get {
                 nil
@@ -65,7 +64,7 @@ public struct NoOpTracingInstrument: TracingInstrument {
         }
 
         public var onEnd: (Span) -> Void {
-            { (span) in
+            { _ in
                 ()
             }
         }
--- a/Sources/Instrumentation/Tracing/NoOpTracing.swift
+++ b/Sources/Instrumentation/Tracing/NoOpTracing.swift
@@ -55,7 +55,7 @@ public struct NoOpTracingInstrument: TracingInstrument {
         public var attributes: SpanAttributes {
             get {
                 [:]
-            } 
+            }
             set {
                 // ignore
             }
--- a/Sources/Instrumentation/Tracing/Span.swift
+++ b/Sources/Instrumentation/Tracing/Span.swift
@@ -24,6 +24,9 @@ public protocol Span {
     /// [OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#span).
     var operationName: String { get }
 
+    /// The status of this span.
+    var status: SpanStatus? { get set }
+
     /// The precise `DispatchTime` of when the `Span` was started.
     var startTimestamp: DispatchTime { get }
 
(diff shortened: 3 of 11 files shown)


Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,46 @@
...
+    func end(at timestamp: DispatchTime)
+}
+
+extension Span {
+    public func end() {
+        self.end(at: .now())
+    }
+}
+
+public struct Event {
```
ktoso: SpanEvent? Event is a very generic word hm

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,46 @@
...
+public protocol Span {
+    var operationName: String { get }
+
+    var startTimestamp: DispatchTime { get }
+    var endTimestamp: DispatchTime? { get }
+
+    var baggage: BaggageContext { get }
+
+    var events: [Event] { get }
+    func addEvent(_ event: Event) -> Self
```
ktoso: `mutating func addEvent(_ event: Event)`

and

`func addingEvent(_ event: Event) -> Self`

the latter can be automatically implemented in extension: 

```
extension Span {
  func addingEvent(_ event: SpanEvent) -> Self { 
    var copy = self
    copy.addEvent(event)
    return copy
  }
}
```

Sources/Instrumentation/Tracing/Span.swift
slashmo: Yep, I'm fine with `SpanEvent`. In OT they have `Event` & `TimedEvent`, but I thought it's not worth the overhead of introducing 2 layers of protocols.

Sources/Instrumentation/Tracing/Span.swift
slashmo: Good one! This aligns well with the Swift naming guidelines 👍

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,94 @@
...
+        copy.addEvent(event)
+        return copy
+    }
+
+    public mutating func end() {
+        self.end(at: .now())
+    }
+}
+
+public struct SpanEvent {
```
ktoso, line 81: Could be `: ExpressibleByStringLiteral`

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,94 @@
...
+    public let name: String
+    public let timestamp: DispatchTime
```
ktoso, line 89: Also 

```suggestion
    /// (Optional) One or more Attributes with the same restrictions as defined for Span Attributes.
    public let attributes: []
    public let timestamp: DispatchTime
```

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+public enum SpanAttribute {
+    case string(String)
+    case int(Int)
+    case double(Double)
+    case bool(Bool)
+    case array([SpanAttribute])
+    case any(CustomStringConvertible)
```
ktoso:
```suggestion
        case stringConvertible(CustomStringConvertible)
```

let's do like that, same as MetadataValue in swift-log.

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,94 @@
...
+public struct SpanEvent {
+    public let name: String
+    public let timestamp: DispatchTime
+
+    public init(name: String, at timestamp: DispatchTime = .now()) {
+        self.name = name
+        self.timestamp = timestamp
+    }
+}
+
```
ktoso, line 102: Mind if we add section marks?

```suggestion
// ==== ----------------------------------------------------------------------------------------------------------------
// MARK: Span Attribute
```

(precisely 120 chars 😉 )

Sources/Instrumentation/Tracing/TracingInstrument.swift
```diff
@@ -0,0 +1,27 @@
...
+
+import Baggage
+import Dispatch
+
+public protocol TracingInstrument: Instrument {
+    var currentSpan: Span? { get }
+
+    func startSpan(named operationName: String, baggage: BaggageContext, at timestamp: DispatchTime) -> Span
+}
+
```
ktoso, line 30: ~are we sure we need the explicit `at timestamp: DispatchTime` param?~ I guess it's ok since we default it hm

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,94 @@
...
+    var baggage: BaggageContext { get }
+
+    var events: [SpanEvent] { get }
+    mutating func addEvent(_ event: SpanEvent)
+
+    var attributes: [String: SpanAttribute] { get }
+    subscript(attributeName attributeName: String) -> SpanAttribute? { get set }
+
+    var onEnd: (Span) -> Void { get }
```
ktoso: This is confusing, why would we enforce this on the protocol? Up to the impl how to handle an end call. :)

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,94 @@
...
+
+public enum SpanAttribute {
+    case string(String)
+    case int(Int)
+    case double(Double)
+    case bool(Bool)
+    case array([SpanAttribute])
+    case any(CustomStringConvertible)
+}
+
```
ktoso, line 125: Should we also offer:

```swift
// MARK: Span Attributes: Semantic Conventions

// https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/span-general.md
extension SpanAttribute {
  // Transport protocol used. See note below.
  public func netTransport(_ value: String) -> SpanAttribute {
    .string(value)
  }
```

ktoso: And offer all those https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/README.md

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,94 @@
...
+    var baggage: BaggageContext { get }
+
+    var events: [SpanEvent] { get }
+    mutating func addEvent(_ event: SpanEvent)
+
+    var attributes: [String: SpanAttribute] { get }
+    subscript(attributeName attributeName: String) -> SpanAttribute? { get set }
+
+    var onEnd: (Span) -> Void { get }
+    mutating func end(at timestamp: DispatchTime)
```
ktoso, line 58: Would you be ok with `finish()` instead?

slashmo: Yep, `finish` is good as well. I followed the OT spec on this one. https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#end

ktoso: Hmm... I see, well that's the spec I guess. Maybe just a `TODO: that we may consider finish?`
End does not sound like a "verb" which this kind of feels it should be 🤔

Sources/Instrumentation/Tracing/TracingInstrument.swift
slashmo: I thought it might be helpful to have an option to supply a timestamp when testing. There's also a write up about this in the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#span-creation

Sources/Instrumentation/Tracing/Span.swift
slashmo: Haha, don't mind at all 😁

Sources/Instrumentation/Tracing/Span.swift
slashmo: Yes, makes sense 👍

Sources/Instrumentation/Tracing/TracingInstrument.swift
ktoso: 👍

ktoso: Should we express it as `DispatchTime?` so it's clearer why context is not last?

and our impl does `timestamp ?? .now()` then

slashmo: Yes, I like that 👍

Sources/Instrumentation/Tracing/Span.swift
slashmo: Good point, I'll add the TODO

```suggestion
    // TODO: naming is defined in the spec, but we may want to consider finish instead as it sounds more like a verb
    mutating func end(at timestamp: DispatchTime)
```

Sources/Instrumentation/Tracing/Span.swift
slashmo: I was already thinking about whether we should bake all of these semantics into the `Instrument` library or make them available separately (domain-specific) instead. What's your opinion about this?

Sources/Instrumentation/Tracing/TracingInstrument.swift
ktoso: Cool excellent

Sources/Instrumentation/Tracing/Span.swift
ktoso: I see, you mean since that's OpenTelemetry, that's a good point. I wonder, technically we are pretty much conforming anyway to OT here, so maybe can do this as well... 

Can do as a separate ticket, could you ticketify?

slashmo: https://github.com/slashmo/gsoc-swift-tracing/issues/65

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    var startTimestamp: DispatchTime { get }
+    var endTimestamp: DispatchTime? { get }
+
+    var baggage: BaggageContext { get }
+
+    var events: [SpanEvent] { get }
+    mutating func addEvent(_ event: SpanEvent)
+
+    var attributes: [String: SpanAttribute] { get }
```
ktoso: > Attributes SHOULD preserve the order in which they're set. Setting an attribute with the same key as an existing attribute SHOULD overwrite the existing attribute's value.

implies we might want to not expose a Dictionary here, we can address this in a separate ticket; There's no good sorted dict today sadly; We could consider a struct to hide the dict.

slashmo: Hiding behind a struct sounds useful, then we could solve this via an implementation detail. Maybe an array of tuples like `HTTPHeaders` where instead of simply appending new tuple values updating the existing one with the same key. Might be a performance hit though.

ktoso: Sometimes scanning through an array can be better than a dictionary -- arrays are great with data locality :) Anyway, I'm more thinking if/when we'd get SortedDict that'd be nice to use here perhaps. It depends on amount of the data and how it's used; It can be good for us to hide what we're using to implement this -- then we can change around and noone will notice.

Tests/InstrumentationTests/TracingInstrumentTests.swift
```diff
@@ -111,6 +126,7 @@ struct OTSpan: Span {
 
     mutating func addEvent(_ event: SpanEvent) {
         self.events.append(event)
+        self.isRecording = true
```
slashmo:
```suggestion
```
Not needed anymore as `isRecording` gets set as part of the `didSet`.

Tests/InstrumentationTests/TracingInstrumentTests.swift
```diff
@@ -0,0 +1,216 @@
...
+    func testPlayground() {
+        let httpServer = FakeHTTPServer(instrument: JaegerTracer()) { baggage, _, client -> FakeHTTPResponse in
+            baggage.logger.info("Make subsequent HTTP request")
+            client.performRequest(baggage, request: FakeHTTPRequest(path: "/test", headers: []))
+            return FakeHTTPResponse(status: 418)
+        }
+
+        httpServer.receive(FakeHTTPRequest(path: "/", headers: []))
+    }
+}
```
ktoso, line 30: ok to keep this until we do the next step where we'll have a tracer implementing the API for the end to end use case

Sources/Instrumentation/Instrument.swift
```diff
@@ -66,3 +66,15 @@ public protocol Instrument {
...
+    public func inject<Carrier, Injector>(_ baggage: BaggageContext, into carrier: inout Carrier, using injector: Injector)
+        where
+        Injector: InjectorProtocol,
+        Carrier == Injector.Carrier {}
+
+    public func extract<Carrier, Extractor>(_ carrier: Carrier, into baggage: inout BaggageContext, using extractor: Extractor)
+        where
+        Extractor: ExtractorProtocol,
+        Carrier == Extractor.Carrier {}
+}
```
ktoso, line 80: 👍

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,224 @@
...
+    /// The read-only collection of events which happened within this `Span`.
+    ///
+    /// `SpanEvent`s might be added via the `addEvent` or `addingEvent` `Span` methods.
+    var events: [SpanEvent] { get }
+
+    /// Add a `SpanEvent` in place.
+    /// - Parameter event: The `SpanEvent` to add to this `Span`.
+    mutating func addEvent(_ event: SpanEvent)
+
+    // TODO: Wrap in a struct to hide collection implementation details.
```
ktoso, line 48: ticketify or do right away please, your call :)

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,224 @@
...
+    /// The attributes describing this `Span`.
+    var attributes: [String: SpanAttribute] { get }
+
+    /// Accesses the `SpanAttribute` with the given name for reading and writing.
+    /// - Parameter attributeName: The name of the attribute used to identify the attribute.
+    /// - Returns:The `SpanAttribute` identified by the given name, or `nil` if it's not present.
+    subscript(attributeName attributeName: String) -> SpanAttribute? { get set }
+
+    /// Returns true if this `Span` is recording information like events, attributes, status, etc.
+    var isRecording: Bool { get }
```
ktoso, line 54: 👍

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -0,0 +1,224 @@
...
+
+    // TODO: naming is defined in the spec, but we may want to consider finish instead as it sounds more like a verb
```
ktoso: 👍

ktoso: maybe ticket?

ktoso (approved):
LGTM, the Links and Kind can be done in a follow up PR I think, more details in next comment

ktoso:
Follow ups to do here:

- `SpanAttributes` type as separate ticket: https://github.com/slashmo/gsoc-swift-tracing/issues/68
- `SpanKind` OT style Spans need SpanKind #67
- Notes on Links: OT style Spans: Links #69 


All those can be separate small tickets, and the Links have two parts to them as well -- not lazy and lazy; The lazy thing let's tackle holisically, how to make contexts CoW etc.

Sources/Instrumentation/Tracing/Span.swift
slashmo: Yes, let's resolve this for now by hiding the dictionary behind a struct. 👍

Sources/Instrumentation/Tracing/Span.swift
slashmo: I'll do the wrapping in this PR, then we can think about implementation details in an upcoming issue.

Sources/Instrumentation/Tracing/Span.swift
slashmo: #70 ✅

slashmo:
```suggestion
```

Sources/Instrumentation/Tracing/NoOpTracing.swift
```diff
@@ -52,15 +52,10 @@ public struct NoOpTracingInstrument: TracingInstrument {
...
-        }
-
-        public subscript(attributeName attributeName: String) -> SpanAttribute? {
+        public var attributes: SpanAttributes {
             get {
-                nil
-            }
-            set {
+                [:]
+            } set {
```
ktoso: weird formatting?

```suggestion
            } 
            set {
```

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -161,6 +153,39 @@ extension SpanAttribute: ExpressibleByArrayLiteral {
     }
 }
 
+/// A collection of `SpanAttribute`s.
+public struct SpanAttributes {
+    private var _attributes = [String: SpanAttribute]()
+
+    /// Accesses the `SpanAttribute` with the given name for reading and writing.
+    /// - Parameter name: The name of the attribute used to identify the attribute.
+    /// - Returns:The `SpanAttribute` identified by the given name, or `nil` if it's not present.
```
ktoso:
```suggestion
    /// - Returns: The `SpanAttribute` identified by the given name, or `nil` if it's not present.
```

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -161,6 +153,39 @@ extension SpanAttribute: ExpressibleByArrayLiteral {
...
+    public var isEmpty: Bool {
+        self._attributes.isEmpty
+    }
+}
+
+extension SpanAttributes: ExpressibleByDictionaryLiteral {
+    public init(dictionaryLiteral elements: (String, SpanAttribute)...) {
+        self._attributes = [String: SpanAttribute](uniqueKeysWithValues: elements)
+    }
+}
```
ktoso, line 187: 👍

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -161,6 +153,39 @@ extension SpanAttribute: ExpressibleByArrayLiteral {
...
+        } set {
+            self._attributes[name] = newValue
+        }
+    }
+
+    /// Calls the given callback for each attribute stored in this collection.
+    /// - Parameter callback: The function to call for each attribute.
+    public func forEachAttribute(_ callback: (String, SpanAttribute) -> Void) {
+        self._attributes.forEach { callback($0.key, $0.1) }
+    }
```
ktoso, line 175: 👍 forEach would have been fine IMO, people don't have to realize they use a wrapper

ktoso: (same in other places actually, I don't think I voiced this on BaggageContext but the same thoughts tbh -- very minor thing tho)

ktoso (approved):
LGTM :) Minor things inline

Sources/Instrumentation/Tracing/NoOpTracing.swift
slashmo: I always used to do it like this as I picked it up somewhere (maybe even the Swift book, but not sure). I'm fine with putting it on the next line too.

Sources/Instrumentation/Tracing/Span.swift
slashmo: True, `attributes.forEachAttribute` sounds a bit repetitive. I'll change here and create an issue for it in the `BaggageContext` repository.

PR #71, opened 2020-07-15
Add SpanKind type

### LINKED ISSUE #67: OT style Spans need SpanKind
See https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#spankind

and https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/trace/Span.java#L530

Note that we set this when creating a new span, Java uses a builder pattern but we'd use a `func startSpan(operationName: String, spanKind: SpanKind = .internal)` I guess

slashmo:
Closes #67

### DIFF
--- a/Sources/Instrumentation/Tracing/Span.swift
+++ b/Sources/Instrumentation/Tracing/Span.swift
@@ -24,6 +24,9 @@ public protocol Span {
     /// [OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#span).
     var operationName: String { get }
 
+    /// The kind of this span.
+    var kind: SpanKind { get }
+
     /// The status of this span.
     var status: SpanStatus? { get set }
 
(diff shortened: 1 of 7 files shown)


Sources/Instrumentation/Tracing/TracingInstrument.swift
```diff
@@ -25,7 +25,12 @@ public protocol TracingInstrument: Instrument {
     ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
     ///   - context: The `BaggageContext` within to start the new `Span`.
     ///   - timestamp: The `DispatchTime` at which to start the new `Span`.
```
ktoso, line 28: docs for the kind and mentionit defaults to .internal

ktoso (approved):
LGTM, one nitpick :)

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -246,3 +249,24 @@ public struct SpanStatus {
...
+    /// This span is the child of a remote `.client` span that was expected to wait for a response.
+    case server
+    /// Indicates that the span describes a synchronous request to some remote service.
+    /// This span is the parent of a remote `.server` span and waits for its response.
+    case client
+    /// Indicates that the span describes the parent of an asynchronous request. This parent span is expected to end before the corresponding child
+    /// `.consumer` span, possibly even before the child span starts. In messaging scenarios with batching,
+    /// tracing individual messages requires a new `.producer` span per message to be created.
+    case producer
+    /// Indicates that the span describes the child of an asynchronous `.producer` request.
```
ktoso, line 268: that's direct wording from OT? It's kind of bad to call those things requests to be honest... it's messages, but let's go with it ;)

slashmo: Yep, copied them over from the OT spec.

Sources/Instrumentation/Tracing/TracingInstrument.swift
slashmo: Good catch 👍 Although I wouldn't mention the default up here. It's documented in the protocol extension where the actual default value is being set.

ktoso: Hmm... in jazzy that's a bit hard to discover AFAIR but maybe you're right -- let's just add the param info

slashmo: I'll check how it gets rendered by [swift-doc](https://github.com/swiftdocorg/swift-doc).

slashmo: It could also be worth considering adding something like `swift-doc` to the CI to ensure complete documentation to catch things like this earlier. Do you think that's worth it?

ktoso: would swift-doc fail on it? That's nice but it can also be annoying to have an overly zealous CI about doc comments which we may have in flux hm.

Your call if you want to include it.

I'm ok with code review being the guard for good docs 😉

slashmo: swift-doc can report the coverage of documentation for public API. I'm also happy with leaving it up to code reviews. 👍

PR #73, opened 2020-07-16
Add SpanLink type

### LINKED ISSUE #69: OTel style Spans: Links
Some notes about how links can be done:

The Java impl does:

```java
    Builder addLink(SpanContext spanContext);
    Builder addLink(SpanContext spanContext, Attributes attributes);
//    * <p>Links are used to link {@link Span}s in different traces. Used (for example) in batching operations, where a single batch handler processes multiple requests from different traces or the same trace.
    Builder addLink(Link link);
```

We can basically look the same here:

```swift
    mutating func addLink(context: BaggageContext, attributes: SpanAttributes = .none)
    mutating func addLink(_ link: Link)
```

Same as with any other `add` things we can add `adding...` ones if we want via extensions later on.

as for the comment on the spec:

> - An API to record a single Link where the Link properties are passed as arguments. This MAY be called AddLink.
> - An API to record a single Link **whose attributes or attribute values are lazily constructed, with the intention of avoiding unnecessary work if a link is unused.** If the language supports overloads then this SHOULD be called AddLink otherwise AddLazyLink MAY be considered.

I think we can handle the lazy case inside the `SpanAttributes` https://github.com/slashmo/gsoc-swift-tracing/issues/68

To be honest this starts to lead us into that these context things should become Copy-on-Write types, so we can share them -- and this is a perfect example since we'll never mutate the span attributes here 🤔 

As step 0 we can just be eager for now, but let's then make a ticket about the lazy part -- I think we should solve it by copy-on-write style things rather than some `() -> SpanAttributes` which would have been the other option.

slashmo:
Closes #69 

Instead of, as suggested in #69, I only added one method called `addLink(_ link: SpanLink)`. I think we won't need an additional `addLink(_ context: BaggageContext, attributes: SpanAttributes)` requirement. What do you think @ktoso?

### DIFF
--- a/Sources/Instrumentation/Tracing/Span.swift
+++ b/Sources/Instrumentation/Tracing/Span.swift
@@ -48,13 +48,18 @@ public protocol Span {
     /// - Parameter event: The `SpanEvent` to add to this `Span`.
     mutating func addEvent(_ event: SpanEvent)
 
-    // TODO: Wrap in a struct to hide collection implementation details.
-
     /// The attributes describing this `Span`.
     var attributes: SpanAttributes { get set }
 
     /// Returns true if this `Span` is recording information like events, attributes, status, etc.
     var isRecording: Bool { get }
+
+    /// The read-only collection of linked `Span`s.
+    var links: [SpanLink] { get }
+
+    /// Add a `SpanLink` in place.
+    /// - Parameter link: The `SpanLink` to add to this `Span`.
+    mutating func addLink(_ link: SpanLink)
 
     /// End this `Span` at the given timestamp.
     /// - Parameter timestamp: The `DispatchTime` at which the span ended.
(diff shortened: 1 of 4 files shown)


Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -48,14 +48,19 @@ public protocol Span {
...
 
     /// Returns true if this `Span` is recording information like events, attributes, status, etc.
     var isRecording: Bool { get }
 
+    /// The read-only collection of linked `Span`s.
+    var links: [SpanLink] { get }
+
+    /// Add a `SpanLink` in place.
+    /// - Parameter link: The `SpanLink` to add to this `Span`.
+    mutating func addLink(_ link: SpanLink)
```
ktoso, line 62: makes sense, would we though want to add extensions for just linking with a baggage context?

```
extension Span { 

  mutating func addLink(_ context: BaggageContext, attributes: SpanAttributes = [:]) { 
    self.addLink(SpanLink(context: context, attributes: attributes))
  }
}
```

Just a nice UX touch :-)

ktoso (approved):
LGTM, can we add the extension too? 👍

Sources/Instrumentation/Tracing/Span.swift
slashmo: Do you have guidance/rules on when to add such extensions? Personally, I'm always very careful not to add too much public API, even if it's in this case just one method.

ktoso: Your gut reaction is a good one 👍 Adding sugar is always to be considered if it's worth it... I'm ok to delay adding :) We can add if it gets annoying 👍 

// Here I kind of felt it'd be nice since the usage site I kind of imagine to be:

```swift
let span = ...

other.onComplete { context in 
    span.addLink(context)
    // continue things
}
```

But I think you may be right... let's not imagine but see in action and then we add it :)

slashmo: Glad to hear that 😊 I also like how your example snippet looks like. Maybe instead we could drop the method that takes a link in favor of adding this method.

ktoso: > Maybe instead we could drop the method that takes a link in favor of adding this method.

That's likely a not so good direction; it's better to err on the "the type" rather than the sugar and "i'll make the type for you", it's more explicit and adding the sugar on top :)

slashmo: Okay, that's a good point 👍

Issue #48, opened 2020-06-23
Share BaggageContext across multiple NIO handlers


slashmo:
Currently, there's no way of sharing `BaggageContext` across multiple NIO channel handlers. This means, each handler would have to create a new `BaggageContext` ([like in this example](https://github.com/slashmo/gsoc-swift-tracing/blob/master/UseCases/Sources/HTTPEndToEnd/Services/OrderServiceHandler.swift#L28)) which eliminates the ability for context propagation across handlers in the same channel pipeline. 

@ktoso & I thought about adding a new `baggage` property to the `ChannelHandlerContext` so that a `BaggageContext` can be created in the outer most channel handler, set on the `Context`, and then be accessed in later channel handlers. This would probably also mean that NIO itself would have to depend on the `BaggageContext` library.

cc @weissi I'd love to get your input on this as well. I'm also happy to to move this over to the NIO repo once we have a general idea on how to make this integration happen 😊

ktoso:
Some extra references:

This would mirror the Netty feature of "attributes", which are:

> Storing stateful information
> 
> AttributeMap.attr(AttributeKey) allow you to store and access stateful information that is related with a handler and its context. Please refer to ChannelHandler to learn various recommended ways to manage stateful information.
> 
> https://netty.io/4.0/api/io/netty/channel/ChannelHandlerContext.html#storing-stateful-information

In Netty a channel handler context `extends AttributeMap` but in our case I think it's more likely that the baggage would be a property inside of the `ChannelHandlerContext` I guess. It's also another case of why we're trying to not use the word context -- to avoid `(nio)context.(baggage)context`.

It would seem this would want to be added to `baggage { get set }` to `ChannelCore`  https://github.com/apple/swift-nio/blob/a91c87f1c297c4326c4bbdec9c0dcb12f32c6a2c/Sources/NIO/Channel.swift#L22 as @weissi mentioned to me some time earlier.

Do we want to also expose it form Channel but "make it automatically hop to the EL if needed"? I'm not sure we need it on channel tbh and on the context could be fine.

Thanks in advance for any hints @weissi and @slashmo hopefully can have a stab at this. 
It would not be merge-ready until the context library is "hardened and ready" so we'd for now be developing things against @slashmo's branch with the context on there. Once we're all happy with things we can figure out how/where the libs live and then it could become "real" in NIO I suppose :)

weissi:
[...]

PR #75, opened 2020-07-19
Remove currentSpan from TracingInstrument API

### LINKED ISSUE #74: Remove tracing instrument current span
Proposing to remove 'tracingInstrument.currentSpan' completely from our APIs 

Thinking some more about it I think we should be doubling down on the fact that we need to pass context around explicitly and let's not try to offer the thread local workaround. If we need to we could revisit this but for now let's remove it and not dive into the world of scope managers at all, as were on the side of explicitly context passing already anyway. 

The main reason is that we want baggage context to be agnostic to any implementation of even tracing and thus everyone should be able to adopt it. And, since there are so many concurrency stories in swift it will be hard or impossible to implement a thread local style current span that actually works reliably - e.g. across nio that then uses dispatch that then uses some couritines etc... let's not assume we are able to solve this for the time being.

slashmo:
Closes #74 

Had a documentation typo anyways 🙈

### DIFF
--- a/Sources/Instrumentation/Tracing/NoOpTracing.swift
+++ b/Sources/Instrumentation/Tracing/NoOpTracing.swift
@@ -16,8 +16,6 @@ import Dispatch
 
 /// No operation TracingInstrument, used when no tracing is required.
 public struct NoOpTracingInstrument: TracingInstrument {
-    public var currentSpan: Span? = nil
-
     public func startSpan(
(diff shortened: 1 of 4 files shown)


ktoso (approved):
LGTM

pokryfka:
Related to https://github.com/slashmo/gsoc-swift-tracing/issues/48

PR #76, opened 2020-07-19
Remove macOS version requirement

slashmo:
This must have been leftover from poking around. I believe there's currently no reason to require a specific macOS version.


ktoso (approved):
Yeah, good. I was wondering why we needed that -- I guess anything that touched vapor or similar libs which do require this

Issue #72, opened 2020-07-16
Test flexibility of TracingInstrument by creating an XRay tracer


slashmo:
Because XRay uses different wordings for things like `Span`s it's a good candidate for testing how well our `TracingInstrument` performs in the real world. We plan on forking [pokryfka/aws-xray-sdk-swift](https://github.com/pokryfka/aws-xray-sdk-swift) to poke around a bit by having it depend on `gsoc-swift-tracing`.

pokryfka:
Cool, let me know if the interface I created is convenient to use and if if you have any comments/issues in general.

While you are more than welcome to fork it and poke around, IMO rather than making `pokryfka/aws-xray-sdk-swift` depend on  `gsoc-swift-tracing` it would be more proper to make another project with dependency on both which would implement `XRayTracingInstrument`. (or maybe target within the project, like with lambda?)

I should probably be as simple as creating a `XRayRecorder` instance within and "feeding" it with whatever comes in with the more generic context.

---

other than different wording there are is also distinction between "annotation" (indexed) and "metadata" - not sure how it compares with OT.

pokryfka:
BTW the main missing/incomplete functionality as far as the `pokryfka/aws-xray-sdk-swift` is concerned is tracing downstream HTTP web services and context propagation which, it seems to me, is one of the main features of `gsoc-swift-tracing`

XRay does support it fairly well but the tracing needs to be "hooked" in somehow.
As far as `swift-server` is concerned `AsyncHTTPClient` seems to be popular choice and would be a convenient place to do it.
However `AsyncHTTPClient` does not have any interface to add "middleware" or tracing, also its not `open` and cannot be extended.

I made a quick hack and implemented [XRayMiddleware](https://github.com/pokryfka/aws-xray-sdk-swift/blob/master/Sources/AWSXRayRecorderSDK/Middleware.swift) for [aws-sdk-swift](https://github.com/swift-aws/aws-sdk-swift) but its broken and I am lurking to see how you address it.

slashmo:
Thanks for chiming in, @pokryfka 👍

I guess we aim for `TracingInstrument` to be the standard way to vend tracing APIs.
[...]

PR #77, opened 2020-07-19
Expose public dictionary init for SpanAttributes

slashmo:
In order to bridge other types (e.g. [`XRayRecorder.Segment.AnnotationValue`](https://github.com/slashmo/aws-xray-sdk-swift/blob/c11535e382be3052525ed9f4eef25f14b3ea89c5/Sources/AWSXRayRecorder/Segment.swift#L360)) we need to expose an additional initializer for `SpanAttributes` that takes a dictionary. Libraries bridging from their own `SpanAttribute`-like type to `SpanAttribute` may add an extension to `SpanAttribute` to bridge a single value, but in order for them to create a `SpanAttributes` (required in the `Span` protocol), it's currently only possible to initialize `SpanAttributes` with a dictionary literal.

### DIFF
--- a/Sources/Instrumentation/Tracing/Span.swift
+++ b/Sources/Instrumentation/Tracing/Span.swift
@@ -165,6 +165,12 @@ extension SpanAttribute: ExpressibleByArrayLiteral {
 public struct SpanAttributes {
     private var _attributes = [String: SpanAttribute]()
 
+    /// Create a set of attributes by wrapping the given dictionary.
+    /// - Parameter attributes: The attributes dictionary to wrap.
+    public init(_ attributes: [String: SpanAttribute]) {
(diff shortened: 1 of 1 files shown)


ktoso (approved):

ktoso:
Have not dug deep if the semantics match, trusting you on this one :)

pokryfka:
actually XRay Annotation cannot have list or dictionary values (only boolean, string and numeric are allowed)

on the other hand Metadata may have list or nested types

see comparison in https://github.com/slashmo/gsoc-swift-tracing/issues/72#issuecomment-660824378

with the current TracingInstrument interface SpanAttribute type should probably be mapped to either annotation or metadata or AWS (environment) or HTTP "attribute" in a very implementation specific way

PR #79, opened 2020-07-20
Make SpanAttribute ExpressibleByStringInterpolation

pokryfka:
metadata/attributes string values are rarely static string literals, more often they contain strings with computed values


ktoso (approved):
Good one, LGTM

PR #81, opened 2020-07-20
!trace #80 adjust overloads on tracing instrumrnt startSpan

### LINKED ISSUE #80: `TracingInstrument` default implementation of startSpan is potentially infinite loop
`TracingInstrument` extends `Instrument` with `startSpan` method:

```swift
/// An `Instrument` with added functionality for distributed tracing. Is uses the span-based tracing model and is
/// based on the OpenTracing/OpenTelemetry spec.
public protocol TracingInstrument: Instrument {
    /// Start a new `Span` within the given `BaggageContext` at a given timestamp.
    /// - Parameters:
    ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
    ///   - context: The `BaggageContext` within to start the new `Span`.
    ///   - kind: The `SpanKind` of the new `Span`.
    ///   - timestamp: The `DispatchTime` at which to start the new `Span`.
    func startSpan(
        named operationName: String,
        context: BaggageContext,
        ofKind kind: SpanKind,
        at timestamp: DispatchTime?
    ) -> Span
}
```

However, at the same time it provides its default implementation with exactly the same signature.

It is possible to create new type of a `TracingInstrument` which does not override the default implementation, example:

```swift
extension XRayRecorder: TracingInstrument {
    public func extract<Carrier, Extractor>(_ carrier: Carrier, into baggage: inout BaggageContext, using extractor: Extractor) where Carrier == Extractor.Carrier, Extractor : ExtractorProtocol {
        // TODO: impl
    }
    
    public func inject<Carrier, Injector>(_ baggage: BaggageContext, into carrier: inout Carrier, using injector: Injector) where Carrier == Injector.Carrier, Injector : InjectorProtocol {
        // TODO: impl
    }
}
```

in which case:

```swift
let instrument: TracingInstrument = XRayRecorder(emitter: emitter)
instrument.startSpan(/ .../)
```

will be recursively calling itself

---

I understand the motivation to provide a default implementation was to provide default value for `SpanKind`.

Perhaps it would be safer if it was optional from the very beginning:

```swift
    func startSpan(
        named operationName: String,
        context: BaggageContext,
        ofKind kind: SpanKind?,
        at timestamp: DispatchTime?
    ) -> Span
```

or not included in the arguments in the extension (in fact the value of `kind` is not used at all even if different than default value):

```swift
    public func startSpan(
        named operationName: String,
        context: BaggageContext,
        at timestamp: DispatchTime? = nil
    ) -> Span {
        self.startSpan(named: operationName, context: context, ofKind: .internal, at: nil)
    }
```

ktoso:
This should address the issue pointed out in #80 

- resolves #80

### DIFF
--- a/Sources/Instrumentation/Tracing/TracingInstrument.swift
+++ b/Sources/Instrumentation/Tracing/TracingInstrument.swift
@@ -22,7 +22,7 @@ public protocol TracingInstrument: Instrument {
     ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
     ///   - context: The `BaggageContext` within to start the new `Span`.
     ///   - kind: The `SpanKind` of the new `Span`.
-    ///   - timestamp: The `DispatchTime` at which to start the new `Span`.
+    ///   - timestamp: The `DispatchTime` at which to start the new `Span`. Passing `nil` leaves the timestamp capture to the underlying tracing instrument.
     func startSpan(
         named operationName: String,
         context: BaggageContext,
@@ -38,11 +38,10 @@ extension TracingInstrument {
     ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
     ///   - context: The `BaggageContext` within to start the new `Span`.
     ///   - kind: The `SpanKind` of the `Span` to be created. Defaults to `.internal`.
-    ///   - timestamp: The `DispatchTime` at which to start the new `Span`. Defaults to `nil`.
+    ///   - timestamp: The `DispatchTime` at which to start the new `Span`. Defaults to `nil`, meaning the timestamp capture is left up to the underlying tracing instrument.
     public func startSpan(
         named operationName: String,
         context: BaggageContext,
-        ofKind kind: SpanKind = .internal,
         at timestamp: DispatchTime? = nil
     ) -> Span {
         self.startSpan(named: operationName, context: context, ofKind: .internal, at: nil)


ktoso:
cc @pokryfka

Sources/Instrumentation/Tracing/TracingInstrument.swift
```diff
@@ -22,7 +22,7 @@ public protocol TracingInstrument: Instrument {
     ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
     ///   - context: The `BaggageContext` within to start the new `Span`.
     ///   - kind: The `SpanKind` of the new `Span`.
-    ///   - timestamp: The `DispatchTime` at which to start the new `Span`.
+    ///   - timestamp: The `DispatchTime` at which to start the new `Span`. Passing `nil` leaves the timestamp capture to the underlying tracing instrument.
```
slashmo, line 25: 💯

Sources/Instrumentation/Tracing/TracingInstrument.swift
```diff
@@ -38,11 +38,10 @@ extension TracingInstrument {
     ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
     ///   - context: The `BaggageContext` within to start the new `Span`.
     ///   - kind: The `SpanKind` of the `Span` to be created. Defaults to `.internal`.
-    ///   - timestamp: The `DispatchTime` at which to start the new `Span`. Defaults to `nil`.
+    ///   - timestamp: The `DispatchTime` at which to start the new `Span`. Defaults to `nil`, meaning the timestamp capture is left up to the underlying tracing instrument.
```
slashmo, line 41: 👌

slashmo (approved):
LGTM, thanks! 👍

PR #82, opened 2020-07-21
!time define Timestamp backed by DispatchWallTime

ktoso:
Fixes two things:

- we used the wrong dispatch time, thanks @pokryfka for spotting this
- it should not be necessary to import Dispatch just to use the tracing instruments; thus introduced a Timestamp type backed by dispatch.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -3,10 +3,10 @@
     "pins": [
       {
         "package": "swift-baggage-context",
-        "repositoryURL": "https://github.com/slashmo/gsoc-swift-baggage-context.git",
+        "repositoryURL": "https://github.com/slashmo/gsoc-swift-baggage-context",
         "state": {
           "branch": null,
-          "revision": "349804a0599573d09df6e433cb8266b80727e5b0",
+          "revision": "10469603494f1cde67fbd0725cdde0e04d131b05",
           "version": "0.1.0"
         }
       },
--- a/Sources/Instrumentation/Tracing/Span.swift
+++ b/Sources/Instrumentation/Tracing/Span.swift
@@ -12,7 +12,6 @@
 //===----------------------------------------------------------------------===//
 
 import Baggage
-import Dispatch
 
 /// A `Span` type that follows the OpenTracing/OpenTelemetry spec. The span itself should not be
 /// initializable via its public interface. `Span` creation should instead go through `tracer.startSpan`
@@ -30,11 +29,11 @@ public protocol Span {
     /// The status of this span.
     var status: SpanStatus? { get set }
 
-    /// The precise `DispatchTime` of when the `Span` was started.
-    var startTimestamp: DispatchTime { get }
+    /// The `Timestamp` of when the `Span` was started.
+    var startTimestamp: Timestamp { get }
 
-    /// The precise `DispatchTime` of when the `Span` has ended.
-    var endTimestamp: DispatchTime? { get }
+    /// The `Timestamp` of when the `Span` has ended.
+    var endTimestamp: Timestamp? { get }
 
     /// The read-only `BaggageContext` of this `Span`, set when starting this `Span`.
     var baggage: BaggageContext { get }
@@ -54,8 +53,8 @@ public protocol Span {
     mutating func addLink(_ link: SpanLink)
 
     /// End this `Span` at the given timestamp.
-    /// - Parameter timestamp: The `DispatchTime` at which the span ended.
-    mutating func end(at timestamp: DispatchTime)
+    /// - Parameter timestamp: The `Timestamp` at which the span ended.
+    mutating func end(at timestamp: Timestamp)
 }
 
 extension Span {
@@ -77,7 +76,7 @@ extension Span {
 // ==== ----------------------------------------------------------------------------------------------------------------
 // MARK: Span Event
 
-/// An event that occured during a `Span`.
+/// An event that occurred during a `Span`.
 public struct SpanEvent {
     /// The human-readable name of this `SpanEvent`.
     public let name: String
@@ -85,15 +84,15 @@ public struct SpanEvent {
     /// One or more `SpanAttribute`s with the same restrictions as defined for `Span` attributes.
     public var attributes: SpanAttributes
 
-    /// The `DispatchTime` at which this event occured.
-    public let timestamp: DispatchTime
+    /// The `Timestamp` at which this event occurred.
+    public let timestamp: Timestamp
 
     /// Create a new `SpanEvent`.
     /// - Parameters:
     ///   - name: The human-readable name of this event.
-    ///   - attributes: The `SpanAttributes` describing this event. Defaults to no attributes.
-    ///   - timestamp: The `DispatchTime` at which this event occured. Defaults to `.now()`.
-    public init(name: String, attributes: SpanAttributes = [:], at timestamp: DispatchTime = .now()) {
+    ///   - attributes: attributes describing this event. Defaults to no attributes.
+    ///   - timestamp: The `Timestamp` at which this event occurred. Defaults to `.now()`.
+    public init(name: String, attributes: SpanAttributes = [:], at timestamp: Timestamp = .now()) {
         self.name = name
         self.attributes = attributes
         self.timestamp = timestamp
@@ -116,7 +115,7 @@ public enum SpanAttribute {
     case double(Double)
     case bool(Bool)
 
-    // TODO: This could be misused to create a heterogenuous array of attributes, which is not allowed in OT:
+    // TODO: This could be misused to create a heterogeneous array of attributes, which is not allowed in OT:
     // https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-attributes
 
     case array([SpanAttribute])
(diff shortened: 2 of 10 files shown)


Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -85,23 +84,23 @@ extension Span {
...
+    /// The `DispatchTime` at which this event occurred.
+    public let timestamp: Timestamp
 
     /// Create a new `SpanEvent`.
     /// - Parameters:
     ///   - name: The human-readable name of this event.
-    ///   - attributes: The `SpanAttributes` describing this event. Defaults to no attributes.
-    ///   - timestamp: The `DispatchTime` at which this event occured. Defaults to `.now()`.
-    public init(name: String, attributes: SpanAttributes = [:], at timestamp: DispatchTime = .now()) {
+    ///   - attributes: attributes describing this event. Defaults to no attributes.
```
slashmo, line 93: Can we add the "The"s back in so it's consistent with the rest? Also happy to change everything else to this style 😊

slashmo (approved):
LGTM, thanks! 👌 Nice idea to create a wrapper type 👍

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -85,23 +84,23 @@ extension Span {
...
+    ///   - timestamp: time at which this event occurred. Defaults to `.now()`.
```
ktoso:
```suggestion
    ///   - timestamp: The `Timestamp` at which this event occurred. Defaults to `.now()`.
```

Sources/Instrumentation/Tracing/Span.swift
ktoso, line 93:
```suggestion
    ///   - attributes: The `SpanAttributes` describing this event. Defaults to no attributes.
```

ktoso:
Applied feedback. It's a bit repetetive but we can do a full sweep of docs some day and decide on style separately :)

pokryfka:
IMO it'd be nice to have a convenience init without `Dispatch` type (nanoseconds or seconds...)

for reference the implementation used in xray sdk https://github.com/pokryfka/aws-xray-sdk-swift/blob/master/Sources/AWSXRayRecorder/Utils/Timestamp.swift

ktoso:
I'm not a fan of having Doubles involved with timestamps to be honest. A convenience init can be added in an internal extension if you really like it.

in case @Lukasa has an opinion here and wants to convience me otherwise about doubles for timestamp init.

pokryfka:
> I'm not a fan of having Doubles involved with timestamps to be honest. A convenience init can be added in an internal extension if you really like it.

agreed, Doubles are not great for timestamps in general (even though Swift Foundation `TimeInterval` *is* Double);

I was thinking more about UInt64 with miliseconds or nanoseconds

I used UInt64 for rawValue and init with Double for testing against Foundation date (Timestamp is an internal type and its an internal initializer) plus in case XRay segment document, the timestamp is serialized as Double anyway;

---

either way - not an issue for me as I do use `DispatchWallTime` internally as well; not sure if it can be taken as a de facto standard and used in public initializers

ktoso:
> either way - not an issue for me as I do use DispatchWallTime internally as well; not sure if it can be taken as a de facto standard and used in public initializers

Yeah that's a valid point... I guess a nanos based UInt64 repr would be the most common denominator; I'll give it another spin tomorrow

pokryfka:
one more reference from [https://github.com/swift-server/swift-aws-lambda-runtime](https://github.com/swift-server/swift-aws-lambda-runtime/blob/master/Sources/AWSLambdaRuntimeCore/Utils.swift)

```swift
internal extension DispatchWallTime {
    init(millisSinceEpoch: Int64) {
        let nanoSinceEpoch = UInt64(millisSinceEpoch) * 1_000_000
        let seconds = UInt64(nanoSinceEpoch / 1_000_000_000)
        let nanoseconds = nanoSinceEpoch - (seconds * 1_000_000_000)
        self.init(timespec: timespec(tv_sec: Int(seconds), tv_nsec: Int(nanoseconds)))
    }

    var millisSinceEpoch: Int64 {
        Int64(bitPattern: self.rawValue) / -1_000_000
    }
}
```

ktoso:
That's a nice init and names 👍 I'll adopt those :-)

slashmo:
@ktoso Shall I add that initializer so we can merge this in?

ktoso:
Added the inits, could you have a look ?

ktoso:
Rebased, please have a look @slashmo

Package.resolved
```diff
@@ -3,10 +3,10 @@
     "pins": [
       {
         "package": "swift-baggage-context",
-        "repositoryURL": "https://github.com/slashmo/gsoc-swift-baggage-context.git",
+        "repositoryURL": "https://github.com/slashmo/gsoc-swift-baggage-context",
         "state": {
           "branch": null,
-          "revision": "349804a0599573d09df6e433cb8266b80727e5b0",
+          "revision": "10469603494f1cde67fbd0725cdde0e04d131b05",
```
slashmo, line 9: Could you please update this to the latest version? I think I force-pushed the tag directly after creation, could have messed it up a bit 😬 Also, do you know if there's a difference in having the `.git` suffix. I think I remember some SwiftPM bug where it has to be `.git` everywhere, but could have already been fixed and I'm still paranoid by it 😁

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -12,7 +12,6 @@
 //===----------------------------------------------------------------------===//
 
 import Baggage
-import Dispatch
```
slashmo, line 15: 👍

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -55,7 +54,7 @@ public protocol Span {
 
     /// End this `Span` at the given timestamp.
     /// - Parameter timestamp: The `DispatchTime` at which the span ended.
```
slashmo:
```suggestion
    /// - Parameter timestamp: The `Timestamp` at which the span ended.
```

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -77,23 +76,23 @@ extension Span {
...
 public struct SpanEvent {
     /// The human-readable name of this `SpanEvent`.
     public let name: String
 
     /// One or more `SpanAttribute`s with the same restrictions as defined for `Span` attributes.
     public var attributes: SpanAttributes
 
-    /// The `DispatchTime` at which this event occured.
-    public let timestamp: DispatchTime
+    /// The `DispatchTime` at which this event occurred.
```
slashmo:
```suggestion
    /// The `Timestamp` at which this event occurred.
```

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -77,23 +76,23 @@ extension Span {
...
+    public let timestamp: Timestamp
 
     /// Create a new `SpanEvent`.
     /// - Parameters:
     ///   - name: The human-readable name of this event.
-    ///   - attributes: The `SpanAttributes` describing this event. Defaults to no attributes.
-    ///   - timestamp: The `DispatchTime` at which this event occured. Defaults to `.now()`.
-    public init(name: String, attributes: SpanAttributes = [:], at timestamp: DispatchTime = .now()) {
+    ///   - attributes: attributes describing this event. Defaults to no attributes.
+    ///   - timestamp: time at which this event occurred. Defaults to `.now()`.
```
slashmo:
```suggestion
    ///   - timestamp: The `Timestamp` at which this event occurred. Defaults to `.now()`.
```

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -116,7 +115,7 @@ public enum SpanAttribute {
     case double(Double)
     case bool(Bool)
 
-    // TODO: This could be misused to create a heterogenuous array of attributes, which is not allowed in OT:
+    // TODO: This could be misused to create a heterogeneous array of attributes, which is not allowed in OT:
```
slashmo, line 118: Oops!...I Did It Again 🙈

Tests/InstrumentationTests/Tracing/TimestampTests.swift
```diff
@@ -0,0 +1,23 @@
...
+// See LICENSE.txt for license information
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Instrumentation
+import XCTest
+
+final class TimestampTests: XCTestCase {
```
slashmo, line 17: 👍

slashmo (approved):
Looks great 👌 Nitpicks about updating `BaggageContext` & some documentation comments inline.

Sources/Instrumentation/Tracing/Span.swift
ktoso: :-) God bless built-in dictionaries.

Package.resolved
ktoso: Thought that's the latest, will double check with an update 👍

Issue #83, opened 2020-07-21
Consider simplifying SpanAttribute types - use Swift types?


pokryfka:
Per [OT](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-attributes):

> The Span interface MUST provide:
> An API to set a single Attribute where the attribute properties are passed as arguments. This MAY be called SetAttribute.
> To avoid extra allocations some implementations may offer a separate API for each of the possible value types.

`TracingInstrument` defines two public type: `SpanAttribute` and `SpanAttributes` which provide A LOT more and need to be implemented by any `TracingInstrument` implementation either by:

- embracing them and adding dependency on the types even if they not directly match their own internal definition of "an attribute" (see XRays annotation and metadata); or
- adding lots of bridging and copying data back and forward in the process

to be honest I think `SpanAttribute` and `SpanAttributes` are a nice example of Swift type safety but its a bit of an hassle to adapt to

in my original XRay Segment implementation I defined only setters (no getters) for XRay Annotation and decided to expose only Swift types:

```swift
    internal typealias Annotations = [String: AnnotationValue]

    internal enum AnnotationValue {
        case string(String)
        case integer(Int)
        case float(Float)
        case bool(Bool)
    }

    private func setAnnotations(_ newElements: Annotations) {
        lock.withWriterLockVoid {
            for (k, v) in newElements {
                _annotations.updateValue(v, forKey: k)
            }
        }
    }

    public func setAnnotation(_ value: String, forKey key: String) {
        setAnnotations([key: .string(value)])
    }
    public func setAnnotation(_ value: Bool, forKey key: String) {
        setAnnotations([key: .bool(value)])
    }

    public func setAnnotation(_ value: Int, forKey key: String) {
        setAnnotations([key: .integer(value)])
    }

    public func setAnnotation(_ value: Float, forKey key: String) {
        setAnnotations([key: .float(value)])
    }
```

added a getter and a function to remove a value trying to conform to `Span` type but that's still not enough,
I need to expose a (copy) of the collection with all the annotations and a function with a callback modifying all the annotations.

Which OT requirement does it fulfils and what is a use case for that?

I think in a typical use case user creates a span, adds a few attributes (perhaps overriding previous ones) and that's it really.

slashmo:
Thanks for this 👍

## SpanAttribute

That's a good point. What if we'd additionally provide methods for accepting the different supported Swift types in a protocol extension?

[...]

PR #92, opened 2020-07-22
Add primitive Span attribute setters

### LINKED ISSUE #87: Make Span attributes write-only
Currently `Span` attributes are modeled through two types `SpanAttributes` & `SpanAttribute`. To make conformance to the `Span` protocol easier we want to drop these types from the public API and instead expose methods on `Span` for the different primitives, e.g. `setAttribute(_ string: String, forKey key: String)` instead of the subscript-based `attributes` property.

slashmo:
Part of #83
Closes #87 

CC @pokryfka

### DIFF
--- a/Sources/TracingInstrumentation/NoOpTracingInstrument.swift
+++ b/Sources/TracingInstrumentation/NoOpTracingInstrument.swift
@@ -54,14 +54,16 @@ public struct NoOpTracingInstrument: TracingInstrument {
 
         public mutating func addEvent(_ event: SpanEvent) {}
 
-        public var attributes: SpanAttributes {
-            get {
-                [:]
-            }
-            set {
-                // ignore
-            }
-        }
+        public mutating func setAttribute(_ value: String, forKey key: String) {}
+        public mutating func setAttribute(_ value: [String], forKey key: String) {}
+        public mutating func setAttribute(_ value: Int, forKey key: String) {}
+        public mutating func setAttribute(_ value: [Int], forKey key: String) {}
+        public mutating func setAttribute(_ value: Double, forKey key: String) {}
+        public mutating func setAttribute(_ value: [Double], forKey key: String) {}
+        public mutating func setAttribute(_ value: Bool, forKey key: String) {}
+        public mutating func setAttribute(_ value: [Bool], forKey key: String) {}
+        public mutating func setAttribute(_ value: CustomStringConvertible, forKey key: String) {}
+        public mutating func setAttribute(_ value: [CustomStringConvertible], forKey key: String) {}
 
         public let isRecording = false
 
--- a/Sources/TracingInstrumentation/Span.swift
+++ b/Sources/TracingInstrumentation/Span.swift
@@ -42,8 +42,65 @@ public protocol Span {
     /// - Parameter event: The `SpanEvent` to add to this `Span`.
     mutating func addEvent(_ event: SpanEvent)
 
-    /// The attributes describing this `Span`.
-    var attributes: SpanAttributes { get set }
+    /// Set a `String` value for the given attribute key.
+    /// - Parameters:
+    ///   - value: The `String` value to set for the key.
+    ///   - key: The attribute key.
+    mutating func setAttribute(_ value: String, forKey key: String)
+
+    /// Set a `[String]`value for the given attribute key.
+    /// - Parameters:
+    ///   - value: The `[String]` value to set for the key.
+    ///   - key: The attribute key.
+    mutating func setAttribute(_ value: [String], forKey key: String)
+
+    /// Set a `Int` value for the given attribute key.
+    /// - Parameters:
+    ///   - value: The `Int` value to set for the key.
+    ///   - key: The attribute key.
+    mutating func setAttribute(_ value: Int, forKey key: String)
+
+    /// Set a `[Int]` value for the given attribute key.
+    /// - Parameters:
+    ///   - value: The `[Int]` value to set for the key.
+    ///   - key: The attribute key.
+    mutating func setAttribute(_ value: [Int], forKey key: String)
+
+    /// Set a `Double` value for the given attribute key.
+    /// - Parameters:
+    ///   - value: The `Double` value to set for the key.
+    ///   - key: The attribute key.
+    mutating func setAttribute(_ value: Double, forKey key: String)
+
+    /// Set a `[Double]` value for the given attribute key.
+    /// - Parameters:
+    ///   - value: The `[Double]` value to set for the key.
+    ///   - key: The attribute key.
+    mutating func setAttribute(_ value: [Double], forKey key: String)
+
+    /// Set a `Bool` value for the given attribute key.
+    /// - Parameters:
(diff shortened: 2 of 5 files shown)


Tests/InstrumentationTests/TracingInstrumentTests.swift
```diff
@@ -114,12 +113,52 @@ struct OTSpan: Span {
...
+        self.attributes[key] = .bool(value)
+    }
+
+    mutating func setAttribute(_ value: [Bool], forKey key: String) {
+        self.attributes[key] = .array(value.map(SpanAttribute.bool))
+    }
+
+    mutating func setAttribute(_ value: CustomStringConvertible, forKey key: String) {
+        self.attributes[key] = .stringConvertible(value)
+    }
```
pokryfka, line 156: its more of a convenience method and does not define new type as such,

personally I'd add it to protocol extension (code for ilustration purposes, did not test it):

```swift
extension Span {
    public func setAnnotation(_ value: CustomStringConvertible, forKey key: String) {
       setAnnotation(String(describing: value), forKey: key)
    }
}
```

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -48,8 +48,65 @@ public protocol Span {
...
+    /// - Parameters:
+    ///   - value: The `CustomStringConvertible` value to set for the key.
+    ///   - key: The attribute key.
+    mutating func setAttribute(_ value: CustomStringConvertible, forKey key: String)
+
+    /// Set a `[CustomStringConvertible]` value for the given attribute key.
+    /// - Parameters:
+    ///   - value: The `[CustomStringConvertible]` value to set for the key.
+    ///   - key: The attribute key.
+    mutating func setAttribute(_ value: [CustomStringConvertible], forKey key: String)
```
pokryfka: its more of a convenience method and does not define new type as such,

personally I'd add it to protocol extension (code for ilustration purposes, did not test it):

```swift
extension Span {
    public func setAnnotation(_ value: CustomStringConvertible, forKey key: String) {
       setAnnotation(String(describing: value), forKey: key)
    }
}
```

pokryfka: BTW I think `ExpressibleByStringInterpolation` is very convinient for all "metadata",
that way you can use:

```swift
let result = true
let durationInSeconds: Float = 1.2 
setAnnotation("result: \(result), duration: \(durationInSeconds)s"), forKey: "key")
```

similarly, protocol extension could add implementation for all `TracingInstruments` without adding more requirements in the protocol itself

pokryfka: last but not least, `@autoclosure` is sometimes handy, example:

```swift
    public func setAnnotation(_ value: @autoclosure () -> String, forKey key: String)
```

 (maybe let's leave it for now as an idea for further refinements/improvements)

ktoso: Yeah maybe it's worth doing all those as autoclosures, same as we did in the logging API.
It may be a bit weird to do that for a simple Int etc tbh but if we do we should do so consistently for all set... funcs.

It would save some allocs when not recording.

ktoso: > `setAnnotation("result: \(result), duration: \(durationInSeconds)s"), forKey: "key")`

that's just a String, no need to use expressible by string interpolation here.

Esp with an autoclosure there's nothing to gain there, we're already delaying rendering the string with the closure.

ktoso (commented):
Should this PR also remove `SpanAttributes`?

Sources/Instrumentation/Tracing/Span.swift
slashmo: That's a great point, thanks @pokryfka I'll add both `CustomStringConvertible` methods in an extension 👍

pokryfka: I think @ktoso is right `ExpressibleByStringInterpolation` should be not needed if argument type is `String` itself,
just see (in unit test?) if `setAttribute("test \(1)", forKey key: "key")` works "out of the box"
`String` uses different initializers for string literals and if it uses string interpolation both should work

slashmo: @pokryfka Not sure what you mean by that. Isn't it clear that when using `String` interpolation the `String` `setAttribute` method is being used?

slashmo:
> Should this PR also remove `SpanAttributes`?

I wasn't sure whether we want to keep it around as a convenience thing in case other tracers want to use the `SpanAttributes` directly. Also, it's still being used in `SpanEvent` & `SpanLink`.

ktoso:
What's the plan here, that we landed "fancy well typed" attributes?

pokryfka:
From my perspective the main point was not if to use or not primitives (although that would help avoiding some mapping) but to be able to create an implementation of a `Span` without a need to implement getters (against OTel) and using/implementing `SpanAttributes` type.

I was trying to make a segment be a span:

```swift
extension XRayRecorder.Segment: TracingInstrument.Span {
}
```

But if that's difficult I can always create a wrapper type:

```swift
internal struct SpanWithSegment: TracingInstrument.Span {
private let segment: XRayRecorder.Segment
var attributes: TracingInstrument.SpanAttributes
}
```

slashmo:
Putting this on hold for now as we want to focus on other things (HTTPClient & gRPC) first.

ktoso:
I worked on this recently in preparation for the repo move.

Long story short: we can't get rid of the getters because of swift limitations if we want to keep the "nice dsl" (and we do).

This is because the nice DSL is enabled by: 

```swift
    public subscript<T>(dynamicMember dynamicMember: KeyPath<SpanAttribute, SpanAttributeKey<T>>) -> SpanAttribute?
```

and Swift today is lacking set-only subscripts. This may change in the future though.

ktoso:
I have addressed the "specific types" question though, after consulting with a few tracer authors, so we'll be doing: 

```
public enum SpanAttribute: Equatable {
    public typealias Key = SpanAttributeKey

    case int(Int64)
    case intArray([Int64])

    case double(Double)
    case doubleArray([Double])

    case bool(Bool)
    case boolArray([Bool])

    case string(String)
    case stringArray([String])

    case stringConvertibleArray([CustomStringConvertible])

```

which don't allow the weird "mixed / nested" attrs -- none of the specs allow for this anyway and only honeycomb actually claimed they support that but in reality didn't because they'd only "..." other values rather than do any special ntesting.

ktoso:
API wise we can go with:

```
    public subscript(_ name: String) -> SpanAttributeConvertible? {
```

and conform types we allow, which allows us to be type-safe in accepting writes; 

Getting values out is awkward but that's okey -- people should not be doing that.

ktoso:
Closing for now -- new repo to be opened soon I hope.
Terribly sorry for the wait on this

pokryfka:
> Closing for now -- new repo to be opened soon I hope.
> Terribly sorry for the wait on this

Thank you for the update for that. Looking forward to stable API and new repo :-)

PR #97, opened 2020-07-26
+poc,instruments PoC of an Instruments.app instrument "tracing" (naive)

ktoso:
Since I had a moment over the holidays here I picked up a PoC i wanted to sanity check since a while.

---

Context: Since a few years Instruments allows for building custom instruments (yeah all the words are terribly overloaded here but I don't think we can do much about that to be honest). See session: https://developer.apple.com/wwdc18/410 Creating Custom Instruments - WWDC 2018

--- 

PoC / "Use Case" of using the tracing instruments API with signposts and collecting in instruments.

This is by no means super end to end perfect, but showcases somewhat the shape and how compatible (or not) the APIs are.

It's fairly okey, the fact that spans are classes means we can easily use them in signpost IDs which is good.

Some comments inline.

---

❗ THIS IS JUST A POC, NOT END GOAL OR "VERY REAL" TRACER ❗

Result:


Sources/Instrumentation/Tracing/NoOpTracing.swift
```diff
@@ -45,7 +45,7 @@ public struct NoOpTracingInstrument: TracingInstrument {
 
         public var endTimestamp: Timestamp?
 
-        public var baggage: BaggageContext {
+        public var context: BaggageContext {
```
ktoso: See below, on Span why this name change is right IMHO

Sources/Instrumentation/Tracing/Span.swift
```diff
@@ -36,7 +36,7 @@ public protocol Span {
     var endTimestamp: Timestamp? { get }
 
     /// The read-only `BaggageContext` of this `Span`, set when starting this `Span`.
-    var baggage: BaggageContext { get }
+    var context: BaggageContext { get }
```
ktoso: IMHO these should be context because span.context rather than span.baggage.

The baggage word is only to be used to disambiguate in cases like context.context so this isn't one of those cases IMO. (rules we arrived at https://github.com/slashmo/gsoc-swift-baggage-context#existing-context-argument )

UseCases/Sources/InstrumentsAppTracing/OSSignpostTracing.swift
```diff
@@ -0,0 +1,267 @@
...
+import Foundation // string conversion for os_log seems to live here
+
+#if os(macOS) || os(tvOS) || os(iOS) || os(watchOS)
+import os.log
+import os.signpost
+
+// ==== ----------------------------------------------------------------------------------------------------------------
+// MARK: OSSignpost Tracing
+
+@available(OSX 10.14, *)
```
ktoso, line 25: since signposts are only available since then.
(I may have gotten the availability wrong on other OSes)

UseCases/Sources/InstrumentsAppTracing/OSSignpostTracing.swift
```diff
@@ -0,0 +1,267 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Baggage
+import Instrumentation
+import Foundation // string conversion for os_log seems to live here
+
+#if os(macOS) || os(tvOS) || os(iOS) || os(watchOS)
```
ktoso, line 18: this is an apple platforms only thing

UseCases/Sources/InstrumentsAppTracing/OSSignpostTracing.swift
```diff
@@ -0,0 +1,267 @@
...
+        if self.endTimestamp == nil {
+            print("""
+                  warning: 
+                  Span \(self.signpostID) (\(self.operationName)) \
+                  [todo:source location] \
+                  was dropped without end() being called!
+                  """)
+        }
+    }
+    #endif
```
ktoso, line 170: very useful pattern IMHO btw.

UseCases/Sources/InstrumentsAppTracing/OSSignpostTracing.swift
```diff
@@ -0,0 +1,267 @@
...
+        }
+        self.endTimestamp = timestamp
+
+        os_signpost(
+            .end,
+            log: self.log,
+            name: self.signpostName,
+            signpostID: self.signpostID,
+            Self.endFormat
+        )
```
ktoso, line 224: Instruments correlates start/end by the signpost ID, which is why we created it with `self` so we use the objects identity as the correlation ID.

UseCases/Sources/InstrumentsAppTracing/OSSignpostTracing.swift
```diff
@@ -0,0 +1,267 @@
...
+// ==== ----------------------------------------------------------------------------------------------------------------
+// MARK: Baggage Keys
+
+@available(OSX 10.14, *)
+@available(iOS 10.0, *)
+@available(tvOS 10.0, *)
+@available(watchOS 3.0, *)
+enum OSSignpostTracingKeys {
+    enum TraceParentIDs: BaggageContextKey {
+        typealias Value = [OSSignpostID]
```
ktoso, line 237: not "really" used and I don't think there's a good way to make use of this info in instruments... though I may have to learn more about it. Not a huge goal of this PoC anyway.

UseCases/Sources/InstrumentsAppTracing/main.swift
```diff
@@ -0,0 +1,57 @@
...
+let context = BaggageContext()
+
+for i in 1...5 {
+    print("Starting operation: op-\(i)")
+    var parentSpan = tracer.startSpan(named: "op-\(i)", context: context)
+    defer { parentSpan.end() }
+
+    DispatchQueue.global().async {
+        var span = tracer.startSpan(named: "op-\(i)-inner", context: .init()) // empty context
+        span.addLink(SpanLink(context: parentSpan.context)) // TODO: span.addLink(parentSpan)
```
ktoso, line 46: TODO: I would really like that sugar to be honest...

UseCases/Sources/InstrumentsAppTracingInstrpkg/SpansExampleInstrument/SpansExampleInstrument/SpansExampleInstrument.instrpkg
```diff
@@ -0,0 +1,105 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<package>
+	<id>org.swift.server.tracing.example.SpansExampleInstrument</id>
+	<title>SpansExampleInstrument</title>
+	<owner>
+		<name>Konrad `ktoso` Malawski</name>
+	</owner>
+
+    <!-- Instruments Developer Help: https://help.apple.com/instruments/developer/mac/current/ -->
```
ktoso, line 9: see here for docs how this format works, and also WWDC sessions on custom instruments:

- https://developer.apple.com/wwdc18/410 
- https://developer.apple.com/wwdc19/421

UseCases/Sources/InstrumentsAppTracingInstrpkg/SpansExampleInstrument/SpansExampleInstrument/SpansExampleInstrument.instrpkg
```diff
@@ -0,0 +1,105 @@
...
+<!--        <column>-->
+<!--            <mnemonic>thread-end</mnemonic>-->
+<!--            <title>Thread (end)</title>-->
+<!--            <type>thred</type>-->
+<!--            <expression>?thread</expression>-->
+<!--        </column>-->
+    </os-signpost-interval-schema>
+
+    <!-- MARK: Instrument Definitions -->
+    <instrument>
```
ktoso, line 65: just a simple instrument that shows each span (grouped by operation name) as it's own row.

This is obviously would not scale all that well in reality, but it's good for the PoC.

ktoso:
Can this be merged, it's just another showcase -- wdyt @slashmo ?

I'll rebase after review.

Sources/Instrumentation/Tracing/Span.swift
pokryfka: requires major version bump ;-)

pokryfka:
I think this would be very useful with NIO channels, to see in Xcode when exactly - including the order- each job is run.

https://github.com/slashmo/gsoc-swift-tracing/issues/48

UseCases/Sources/InstrumentsAppTracing/OSSignpostTracing.swift
```diff
@@ -0,0 +1,267 @@
...
+        OSSignpostSpan(
+            log: self.log,
+            named: operationName,
+            signpostName: self.signpostName,
+            context: context
+            // , kind ignored
+            // , timestamp ignored, we capture it automatically
+        )
+    }
+}
```
pokryfka, line 67: note that some tracer may choose not to expose setting the timestamp explicitly by user.

this is the case with XRay tracer as well, at the moment (I dont really see a use case for that);
arguably its easier to change XRay tracer than `OSSignpostSpan` ;-)

BTW this is not required by OT

Sources/Instrumentation/Tracing/Span.swift
slashmo: Agreed 👍

Tests/InstrumentationTests/SpanTests.swift
```diff
@@ -17,8 +17,8 @@ import XCTest
 
 final class SpanTests: XCTestCase {
     func testAddingEventCreatesCopy() {
-        // TODO: We should probably replace OTSpan at some point with a NoOpSpan for testing things like this.
-        let span = OTSpan(
+        // TODO: We should probably replace OTelSpan at some point with a NoOpSpan for testing things like this.
+        let span = OTelSpan(
```
slashmo: // OTel all the things 👍

UseCases/Sources/InstrumentsAppTracing/main.swift
slashmo: Would you mind creating an issue for this? I'm also in favor of adding this 🍭

UseCases/Sources/InstrumentsAppTracingInstrpkg/SpansExampleInstrument/SpansExampleInstrument.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,166 @@
+// !$*UTF8*$!
```
slashmo, line 1: I guess we need the Xcode project files for Signpost, right?

slashmo (approved):
Great job with this. I just ran the Instrument and it looks very nice already 👍

slashmo:
> Can this be merged, it's just another showcase

@ktoso Yep, sorry for the delay

UseCases/Sources/InstrumentsAppTracingInstrpkg/SpansExampleInstrument/SpansExampleInstrument.xcodeproj/project.pbxproj
ktoso: For the signpost APIs no, but for the `instrpkg` definition yes.

Tests/InstrumentationTests/SpanTests.swift
ktoso: good :)

Sources/Instrumentation/Tracing/Span.swift
ktoso: The project was never tagged so we're not even a 0.0.0, a major bump from nothing still is nothing 😉

UseCases/Sources/InstrumentsAppTracing/main.swift
ktoso: https://github.com/slashmo/gsoc-swift-tracing/issues/108 and done now. 👍

ktoso:
Thanks folks, just another fun example to include in the suite, merged then :shipit:

PR #100, opened 2020-07-29
Move TracingInstrument & related types to separate library

### LINKED ISSUE #99: Solve "getting" (specific) tracers from instrumentation system
We generally allow to bootstrap a system with an instrument, or a multiplex one.

We should make it possible to:

```swift
import MyTracing 

 // libs do this:
let whatever = instrumentation.instrument

// end users may want to do this:
let myTracer = instrumentation.myTracer // : MyTracer or crash if not installed
```

We likely should change how we store instruments and have them identified somehow with IDs perhaps?
We should return the multiplexed one when people use "give me whatever instrument" because that's _libraries_ and they should use all instruments available to inject/extract.

slashmo:
- `TracingInstrument` & related types now live in their own library target `TracingInstrumentation`
- `InstrumentationSystem` is now unaware of a specific `Instrument` implementation like `TracingInstrumentation`
- `InstrumentationSystem.instance(of: T.self)` allows for casting the bootstrapped `Instrument` with a reader-lock

Closes #99


ktoso (commented):
Looks ok in general; could you add some tests how this works in reality, so...

- what happens when we add a few instruments (multiplex) and one of them is a tracer
  - today we'll crash I guess
  - we'd want to be able to install a few instruments and one of them is a tracer, when generic instrument is invoked all should be returned, when a tracer is asked for i guess we ask for the specific one

Imagine an API where there's some baggage instruments that always attach "user id" to all requests, somehow, and there's a tracer as well. The "user id" somehow magically gets attached for resource attribution so thanks to it we can know "aha, all these requests are caused by this user, it's causing overload! let's reject requests from this user for now" while the tracing just does it's own thing as usual. (Just a silly example, but we need to be able to survive this).

I'm thinking that perhaps we need to:

- store a list of instruments
- return a multiplex instrument "on demand" when `.instrument` is used
- return the tracer when a tracer is requestsed...?

Open to better ideas to be honest.

slashmo:
> * store a list of instruments
> * return a multiplex instrument "on demand" when `.instrument` is used
> * return the tracer when a tracer is requestsed...?

Thanks, makes sense. I forgot about `MultiplexInstrument` in the `instance` method. I'm also totally open to better ways of handling this and naming things.

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -58,13 +50,20 @@ public enum InstrumentationSystem {
...
+    /// - Parameter instrumentType: The type of `Instrument` you want to retrieve an instance for.
+    /// - Returns: An `Instrument` instance of the given type or `nil` if no `Instrument` of that type has been bootstrapped.
+    public static func instance<I>(of instrumentType: I.Type) -> I? {
         self.lock.withReaderLock {
-            let tracer: TracingInstrument? = self._tracer
-            let res: TracingInstrument = tracer ?? NoOpTracingInstrument()
-            return res
+            if let multiplexInstrument = self._instrument as? MultiplexInstrument {
+                return multiplexInstrument.firstInstance(of: I.self)
+            }
```
ktoso, line 93: I see, interesting actually hm hm...

Tests/InstrumentationTests/InstrumentationSystemTests.swift
```diff
@@ -12,26 +12,43 @@
...
+        InstrumentationSystem.bootstrapInternal(multiplexInstrument)
+        XCTAssert(InstrumentationSystem.instrument is MultiplexInstrument)
+        XCTAssert(InstrumentationSystem.instance(of: FakeTracer.self) === tracer)
+        XCTAssert(InstrumentationSystem.instance(of: FakeInstrument.self) === instrument)
+
+        InstrumentationSystem.bootstrapInternal(tracer)
+        XCTAssertFalse(InstrumentationSystem.instrument is MultiplexInstrument)
+        XCTAssert(InstrumentationSystem.instance(of: FakeTracer.self) === tracer)
+        XCTAssertNil(InstrumentationSystem.instance(of: FakeInstrument.self))
     }
```
ktoso, line 36: i see that's pretty good actually

ktoso: how about calling it `instrument(of:` tbh? And we'll need also `.instrument` and `tracingInstrument` as convenience 

 so `instrument` gets the whatever was set



tracing instrument is an implicit first where is `TracingInstrument`; it can be an extension in the TracingInstrument library

slashmo: +1 `.tracer` is already implemented this way in `TracingInstrument`, but I'll rename it to `tracingInstrument`.

Sources/Instrumentation/MultiplexInstrument.swift
```diff
@@ -27,6 +27,12 @@ public struct MultiplexInstrument {
     }
 }
 
+extension MultiplexInstrument {
+    func firstInstance<I>(of instrument: I.Type) -> I? {
```
ktoso:
```suggestion
    func firstInstance<I>(of instrument: I.Type) -> I? where I: Instrument {
```

?

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -58,13 +62,20 @@ public enum InstrumentationSystem {
...
 
-    // FIXME: smarter impl
-    public static var tracer: TracingInstrument {
+extension InstrumentationSystem {
+    /// Get an `Instrument` instance of the given type.
+    ///
+    /// When using `MultiplexInstrument`, this returns the first instance of the given type stored in the `MultiplexInstrument`.
+    /// - Parameter instrumentType: The type of `Instrument` you want to retrieve an instance for.
+    /// - Returns: An `Instrument` instance of the given type or `nil` if no `Instrument` of that type has been bootstrapped.
+    public static func instrument<I>(of instrumentType: I.Type) -> I? {
```
ktoso: not just any random `I`, but at least an instrument:

```suggestion
    public static func instrument<I>(of instrumentType: I.Type) -> I? where I: Instrument {
```

right?

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -58,13 +62,20 @@ public enum InstrumentationSystem {
...
         self.lock.withReaderLock { self._instrument }
     }
+}
 
-    // FIXME: smarter impl
-    public static var tracer: TracingInstrument {
+extension InstrumentationSystem {
+    /// Get an `Instrument` instance of the given type.
+    ///
+    /// When using `MultiplexInstrument`, this returns the first instance of the given type stored in the `MultiplexInstrument`.
```
ktoso, line 73: good

Sources/Instrumentation/MultiplexInstrument.swift
```diff
@@ -27,6 +27,12 @@ public struct MultiplexInstrument {
...
+        self.instruments.first(where: { $0 is I }) as? I
```
ktoso: Random sidenote: I REALLY miss Scala's `collect`:

It would have been `self.instruments.iterator.collect { case instrument: I => instrument }.first`

Maybe we'll get nicer partial functions some day... :)

Package.swift
```diff
@@ -52,6 +53,22 @@ let package = Package(
             ]
         ),
 
+        .target(
+            name: "TracingInstrumentation",
+            dependencies: [
+                .product(name: "Baggage", package: "swift-baggage-context"),
+                "Instrumentation"
```
ktoso, line 59: ok; do we need to spell out Baggage dependenccy explicitly tho?

Tests/TracingInstrumentationTests/TracedLock.swift
```diff
@@ -29,7 +30,7 @@ final class TracedLock {
     func lock(context: BaggageContext) {
         // time here
         self.underlyingLock.lock()
-        self.activeSpan = InstrumentationSystem.tracer.startSpan(named: self.name, context: context)
+        self.activeSpan = InstrumentationSystem.tracingInstrument.startSpan(named: self.name, context: context)
```
ktoso, line 33: good, thx

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -14,13 +14,24 @@
...
+/// # Access the Instrument
+/// There are two ways of getting the bootstrapped instrument.
+/// 1. `InstrumentationSystem.instrument`: Returns whatever you passed to `.bootstrap` as an `Instrument`.
+/// 2. `InstrumentationSystem.instrument(of: MyInstrument.self)`: Returns the bootstrapped `Instrument` if it's
+/// an instance of the given type or the first instance of `MyInstrument` if it's part of a `MultiplexInstrument`.
+///
+/// ## What getter to use
+/// - Default to using `InstrumentationSystem.instrument`
+/// - Use `InstrumentationSystem.instrument(of: MyInstrument.self)` only if you need to use specific `MyInstrument` APIs
+/// - Protocols that "inherit" from `Instrument` may also wrap `.instrument(of: TheirInstrument.self)` in a convenience method
```
ktoso: Can you explain a bit more what you mean here?

Do we mean here that libs should provide their own `extension InstrumentationSystem { var coolTracer }`? If so I didn't  understannd it this way hmm

ktoso (commented):
LGTM overall, minor nitpicks inline

ktoso (approved):

Sources/Instrumentation/InstrumentationSystem.swift
slashmo: Yes, wanted to describe exactly that. Did you understand something else instead?

Sources/Instrumentation/InstrumentationSystem.swift
slashmo: I did this at first but had to drop it because of:
> Value of protocol type 'TracingInstrument' cannot conform to 'Instrument'; only struct/enum/class types can conform to protocols

Sources/Instrumentation/MultiplexInstrument.swift
slashmo: *See related comment below*

Package.swift
slashmo: Nope, will remove 👍

Sources/Instrumentation/InstrumentationSystem.swift
ktoso: After re reading it a few times I see what it desccribes but yeah it's a bit confusingly worded, maybe we can improve a but annd make it more example-ish.

```suggestion
/// - Specific instrumentation libraries may also provide their own accessors as extensions, e.g. GreatInstrumentation could provide a `InstrumentationSystem.great` convenience accessor, so prefer using them if available
```

slashmo: Looks good 👍 I'd like to add a sentence for implementors to note that they should use `instrument(of:)` when implementing such convenience methods. What do you think?

```suggestion
/// - Specific instrumentation libraries may also provide their own accessors as extensions, e.g. GreatInstrumentation could provide a `InstrumentationSystem.great` convenience accessor, so prefer using them if available. These accessors should call `.instrument(of: GreatInstrument.self)` under the hood to ensure they work when using a `MultiplexInstrument` as well.
```

Package.swift
ktoso: thanks

Sources/Instrumentation/InstrumentationSystem.swift
ktoso: can you share more context? We can't write what then; `of: TracingInstrument.self`?

if so, we could do another overload for T: TracingInstrument?

Sources/Instrumentation/InstrumentationSystem.swift
ktoso: Sounds good 👍

Sources/Instrumentation/InstrumentationSystem.swift
slashmo: #### `instrument<I: Instrument>(of:I)`

```swift
extension InstrumentationSystem {
    public static func instrument<I>(of instrumentType: I.Type) -> I? where I: Instrument {
        self.lock.withReaderLock {
            if let multiplexInstrument = self._instrument as? MultiplexInstrument {
                return multiplexInstrument.firstInstance(of: I.self)
            }
            return self._instrument as? I
        }
    }
}
```

#### `InstrumentationSystem.tracingInstrument`

```swift
extension InstrumentationSystem {
    public static var tracingInstrument: TracingInstrument {
        instrument(of: TracingInstrument.self)!
    }
}
```

> Value of protocol type 'TracingInstrument' cannot conform to 'Instrument'; only struct/enum/class types can conform to protocols

> if so, we could do another overload for T: TracingInstrument?

Would this be possible in the `TracingInstrumentation` library without exposing `.lock`?

PR #101, opened 2020-07-30
allow for tracingInstrument and tracingInstrument(of: Type) APIs

ktoso:


Sources/TracingInstrumentation/InstrumentationSystem+Tracing.swift
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    /// Returns the `TracingInstrument` bootstrapped as part of the `InstrumentationSystem`.
+    ///
+    /// If the system was bootstrapped with a `MultiplexInstrument` this function attempts to locate the _first_
+    /// tracing instrument as passed to the multiplex instrument. If none is found, a `NoOpTracingInstrument` is returned.
+    ///
+    /// - Returns: A `TracingInstrument` if the system was bootstrapped with one, and `NoOpTracingInstrument` otherwise.
+    public static var tracingInstrument: TracingInstrument {
+        (self.__findInstrument(where: { $0 is TracingInstrument }) as? TracingInstrument) ?? NoOpTracingInstrument()
```
ktoso: to be honest it could cache the value as well btw... but this is ok for now

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -61,24 +61,36 @@ public enum InstrumentationSystem {
...
-    public static func instrument<I>(of instrumentType: I.Type) -> I? {
+    public static func instrument<I>(of instrumentType: I.Type) -> I? where I: Instrument {
+        self.__findInstrument(where: { $0 is I }) as? I
+    }
+
+}
+
+extension InstrumentationSystem {
+    /// :nodoc: INTERNAL API: Do Not Use
+    public static func __findInstrument(where predicate: (Instrument) -> Bool) -> Instrument? {
```
slashmo: Interesting. I've never seen `__` before, only `_`. Is that some kind of convention for public API that shouldn't be used outside the library?

Sources/Instrumentation/MultiplexInstrument.swift
```diff
@@ -28,8 +28,8 @@ public struct MultiplexInstrument {
 }
 
 extension MultiplexInstrument {
-    func firstInstance<I>(of instrument: I.Type) -> I? {
-        self.instruments.first(where: { $0 is I }) as? I
+    func firstInstrument(where predicate: (Instrument) -> Bool) -> Instrument? {
```
slashmo, line 31: 👍

Sources/TracingInstrumentation/InstrumentationSystem+Tracing.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Baggage
+import Instrumentation
+
+
+extension InstrumentationSystem {
+
+    /// Get an `TracingInstrument` instance of the given type.
```
slashmo: On the iPad, so no suggestion: "an" should be "a".

Tests/TracingInstrumentationTests/TracingInstrumentTests.swift
```diff
@@ -27,6 +27,30 @@ final class TracingInstrumentTests: XCTestCase {
 
         httpServer.receive(FakeHTTPRequest(path: "/", headers: []))
     }
+
+    func testItProvidesAccessToATracingInstrument() {
```
slashmo, line 31: Great, thanks for adding a test case for this as well 👌

slashmo (approved):
Looks good. I usually try to avoid exposing "non-public" public API like hell, but in this case it seems like a necessary trait-off.

Sources/TracingInstrumentation/InstrumentationSystem+Tracing.swift
ktoso: thx

ktoso:
```suggestion
    /// Get a `TracingInstrument` instance of the given type.
```

Sources/Instrumentation/InstrumentationSystem.swift
ktoso: It could just be my convention but yea it's like "don't ever even show it in any completion and even think about using this" the :nodoc: is for jazzy to not document the API at all even tho it's public

If you remember LinuxMain those also used `public func __allTests() -> [XCTestCaseEntry] {`

ktoso: Let me confirm with some folks on guidelines...

ktoso: Got some feedback that `_` is enough to hide from autocompletion so changed that

slashmo: Okay 👍

ktoso:
Applied feedback, ready to merge (into your PR)

PR #103, opened 2020-07-30
Add semantic span attributes

### LINKED ISSUE #65: Add OT semantic convention extensions to SpanAttribute
As our `Span` & `SpanAttribute` are very much inspired by the OpenTelemetry spec, we might consider adding their semantic conventions as extensions to the `Instrumentation` library as well: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/span-general.md

https://github.com/slashmo/gsoc-swift-tracing/pull/62#discussion_r451946892

slashmo:
Closes #65


slashmo:
@ktoso I added the basic shape of semantic attributes, but I currently only have the HTTP status code. I'd say let's nail down the API first and then add others in as well.

slashmo:
@ktoso What do you think about depending on `NIOHTTP1` in this library? Makes for some nice API like typed `HTTPMethod`s, but also comes with some *baggage*.

ktoso:
> What do you think about depending on NIOHTTP1 in this library? Makes for some nice API like typed HTTPMethods, but also comes with some baggage.

I understand the convenience but let's _not_ at least not in here; there could be a separate lib which does that, but let's not depend on things in here (I know we do today, but I think we should clean all that up eventually. There could be a `OpenTelemetryNIOInstrumentationSupport` lib which adds that you do here with the NIO's types. We may want to version it separately... 🤔 I don't want us to entangle projects too much; let's keep it simple for now

Package.swift
```diff
@@ -74,14 +74,29 @@ let package = Package(
...
             dependencies: [
                 "NIOInstrumentation",
             ]
         ),
 
+        .target(
+            name: "OpenTelemetryInstrumentationSupport",
+            dependencies: [
+                .target(name: "TracingInstrumentation"),
+                .product(name: "NIOHTTP1", package: "swift-nio")
```
ktoso: let's not for now please, just primitives (int string etc).

For example the Lambda library actually hides the NIO types, so it could not use these directly (!); we need to play nice with everyone, so let's use just the primitive types. There can be other libraries in the future which indeed pull in NIO and others to make these attributes even easier to set.

Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+OpenTelemetry.swift
```diff
@@ -0,0 +1,48 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import NIOHTTP1
+import TracingInstrumentation
+
+public protocol SpanAttributeKey {
+    static var name: String { get }
+
+    associatedtype Value: SpanAttributeConvertible
```
ktoso: nice 👍

Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+OpenTelemetry.swift
```diff
@@ -0,0 +1,48 @@
...
+        self.attributes[Key.name] = value.toSpanAttribute()
+    }
+}
+
+extension SpanAttribute {
+    public enum HTTP {
+        public enum Method: SpanAttributeKey {
+            public static let name = "http.method"
+            public typealias Value = HTTPMethod
+        }
```
ktoso: (nice, but let's do it without the NIO dep, so just String) (there could be another lib whicch does the NIO dance)

Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+OpenTelemetry.swift
```diff
@@ -0,0 +1,48 @@
...
+
+        public enum StatusCode: SpanAttributeKey {
+            public static let name = "http.status_code"
+            public typealias Value = Int
+        }
+    }
```
ktoso: nice 👍

Sources/TracingInstrumentation/Span.swift
```diff
@@ -108,6 +108,16 @@ extension SpanEvent: ExpressibleByStringLiteral {
...
 
+public protocol SpanAttributeConvertible {
+    func toSpanAttribute() -> SpanAttribute
+}
+
+extension Int: SpanAttributeConvertible {
+    public func toSpanAttribute() -> SpanAttribute {
+        .int(self)
+    }
+}
```
ktoso, line 165: 👍 (I assume you'll add String int and Array of those :))

slashmo:
> I don't want us to entangle projects too much; let's keep it simple for now

Okay, I totally get that. Sounds good 👌 Would you define our own `HTTPMethod` or simply do it `String`-based then? I think `String`-based is fine as most users probably have a `NIOHTTP1.HTTPMethod` on hand of which they could simply pass along the `rawValue`.

Tests/OpenTelemetryInstrumentationSupportTests/SpanAttributeOpenTelemetryTests.swift
```diff
@@ -0,0 +1,98 @@
...
+import OpenTelemetryInstrumentationSupport
+import TracingInstrumentation
+import XCTest
+
+final class SpanAttributeOpenTelemetryTests: XCTestCase {
+    func testSemanticAttributes() {
+        InstrumentationSystem.bootstrap(FakeTracer())
+        var span = InstrumentationSystem.tracingInstrument.startSpan(named: "test", context: BaggageContext())
+
+        span.setAttribute(418, forKey: SpanAttribute.HTTP.StatusCode.self)
```
ktoso: so the typed keys are great.

I'd like to see if we can improve two things here:
1) if we can figure out a way to write these as `.HTTP.statusCode`
2) avoiding the `.self` there, it's a bit annoying.

2 we can solve by:

```swift

protocol AttrKey {
    associatedtype Value
}

enum Attr: AttrKey {
    typealias Value = Never
    enum HTTP {
        static let statusCode = StatusCode()
        struct StatusCode: AttrKey {
            typealias Value = Int
        }
    }
}

func take<K>(_ v: K.Value, forKey key: K) where K: AttrKey {}

func test() {
    take(200, forKey: Attr.HTTP.statusCode)
    take("nein", forKey: Attr.HTTP.statusCode)
}

```

for 1 I'm not sure...

ktoso:
I think we may be able to do something even more awesome here...


```
struct SpanAttributeKey<T> {}
protocol SpanAttributeNamespace {
}
enum SpanAttribute {
    case __magic
}
extension SpanAttribute {
     var name: SpanAttributeKey<String> { return .init() }
}
extension SpanAttribute {
    var http: HTTPNamespace { HTTPNamespace() }
    struct HTTPNamespace: SpanAttributeNamespace {
        var statusCode: SpanAttributeKey<Int> { return .init() }
    }
}
@dynamicMemberLookup
class CTXAttributes {
    var storage: [String: String] = [:]
    init() {}
//    func setAttribute<T>(value: T, forKey keyPath: KeyPath<SpanAttribute, SpanAttributeKey<T>>) {
//      let k = SpanAttribute.__magic[keyPath: keyPath]
//      print("Setting \(value) for \(k)")
//    }
//    func setAttribute<T>(_ keyPath: KeyPath<SpanAttribute.Type, SpanAttributeKey<T>>, _ value: T) {
//        let k = SpanAttribute.__magic[keyPath: keyPath]
//        print("Setting \(value) for \(k)")
//    }
    subscript<T>(dynamicMember dynamicMember: KeyPath<SpanAttribute, SpanAttributeKey<T>>) -> T? {
        get {
            let k = SpanAttribute.__magic[keyPath: dynamicMember]
            return storage["\(k)"] as? T
        }
        set {
            let k = SpanAttribute.__magic[keyPath: dynamicMember]
            print("Setting \(k) = \(newValue)")
        }
    }
}
func test() {
    let ctx = CTXAttributes()
    ctx.name = "x"
}
```

but this is still incomplete... This is pretty intense code and I had a lot of help from the compiler team to get to this, so if you'd rather have me polish it up to the "shape" tomorrow morning I can do that.

slashmo:
@ktoso The tests are crashing on CI because the `namespace`-case is used as a protocol witness which is a feature introduced in Swift 5.3 🤔 https://github.com/apple/swift-evolution/blob/master/proposals/0280-enum-cases-as-protocol-witnesses.md

ktoso:
We can surround that protocols func/var there in an #if compiler(>=5.3) so it’d help people do the right thing on 5.3...? Wdyt?

PR #104, opened 2020-07-31
+attributes implement type-safe extensible property style attributes

ktoso:
This allows us to build entire type-safe (nested) keyspaces with documentation and the right types for attributes;
We can thus implement a separate module with all the keys defined: 

- https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/trace/semantic_conventions

Maybe `OpenTelemetryAttributes` or just `OpenTelemetryInstrumentationSupport` or some other name...

We can then take it another level by adding `NIO...` (the same) which will allow setting `attribute.http.requestType = nioRequest` etc etc... Really great type-safe sugar for our end users.

Let's _just_ define the non NIO ones for now though.

This PR needs more love, could you take it from here @slashmo ?
We can absolutely go over how this works in a mini call later on today.

Special thanks to @mikeash and @jckarter <3

### DIFF
--- a/Sources/TracingInstrumentation/Span.swift
+++ b/Sources/TracingInstrumentation/Span.swift
@@ -16,11 +16,10 @@ import Baggage
 /// A `Span` type that follows the OpenTracing/OpenTelemetry spec. The span itself should not be
 /// initializable via its public interface. `Span` creation should instead go through `tracer.startSpan`
 /// where `tracer` conforms to `TracingInstrument`.
+///
+/// - SeeAlso: [OpenTelemetry Specification: Span](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#span).
 public protocol Span {
     /// The operation name is a human-readable string which concisely identifies the work represented by the `Span`.
-    ///
-    /// For guideline on how to name `Span`s, please take a look at the
-    /// [OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#span).
     var operationName: String { get }
 
     /// The kind of this span.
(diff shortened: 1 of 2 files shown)


Tests/TracingInstrumentationTests/SpanTests.swift
```diff
@@ -105,17 +105,38 @@ final class SpanTests: XCTestCase {
...
-            return
-        }
+        XCTAssertEqual(attributes["thing.name"], SpanAttribute.string("hello"))
+        XCTAssertEqual(attributes["meaning.of.life"], SpanAttribute.int(42))
+        XCTAssertEqual(attributes["alive"], SpanAttribute.bool(false))
+
+        // An import like: `import OpenTelemetry...` can enable type-safe well defined attributes,
+        // e.g. as defined in https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/trace/semantic_conventions
+        attributes.name = "kappa"
+        attributes.sampleHttp.statusCode = 200
```
ktoso, line 136: This is _awesome_ 🤯 😎

Tests/TracingInstrumentationTests/SpanTests.swift
```diff
@@ -132,3 +153,50 @@ final class SpanTests: XCTestCase {
...
+public struct HTTPAttributes: SpanAttributeNamespace {
+    public var attributes: SpanAttributes
+    public init(attributes: SpanAttributes) {
+        self.attributes = attributes
+    }
+
+    public enum NestedAttributes: NestedSpanAttributesProtocol {
+        case namespace
+        public var statusCode: SpanAttributeKey<Int> {
+            "http.status_code"
```
ktoso, line 187: we can customize the names (because we need to)

Tests/TracingInstrumentationTests/SpanTests.swift
```diff
@@ -132,3 +153,50 @@ final class SpanTests: XCTestCase {
...
+// MARK: Example Span attributes
+
+extension SpanAttribute {
+    var name: SpanAttributeKey<String> {
+        "name"
+    }
+}
+
+extension SpanAttributes {
+    public var sampleHttp: HTTPAttributes {
```
ktoso, line 167: we can provide nested namespaces

Sources/TracingInstrumentation/Span.swift
```diff
@@ -108,24 +107,55 @@ extension SpanEvent: ExpressibleByStringLiteral {
 // ==== ----------------------------------------------------------------------------------------------------------------
 // MARK: Span Attribute
 
-public protocol SpanAttributeConvertible {
```
ktoso, line 111: I just moved them, we still want these 👍

Tests/TracingInstrumentationTests/SpanTests.swift
slashmo: 🤯

slashmo (approved):
This is awesome! 🤩 I'll merge it into my PR, clean it up and fix the tests, so it's ready for *main*-time 👍 Thanks!

Tests/TracingInstrumentationTests/SpanTests.swift
ktoso: And IDEs understand this, even showing docs etc... <3

Sources/TracingInstrumentation/Span.swift
```diff
@@ -136,6 +166,45 @@ public enum SpanAttribute {
...
+        case (.bool(let l), .bool(let r)): return l == r
+        case (.array(let l), .array(let r)): return l == r
+        case (.stringConvertible(let l), .stringConvertible(let r)): return "\(l)" == "\(r)"
+        case (.string, _),
+             (.int, _),
+             (.double, _),
+             (.bool, _),
+             (.array, _),
+             (.stringConvertible, _),
+             (.__magic, _):
```
ktoso, line 186: I'm not sure if we can make this happen without the magic value... 🤔

Tests/TracingInstrumentationTests/SpanTests.swift
```diff
@@ -132,3 +153,50 @@ final class SpanTests: XCTestCase {
...
+    }
+
+    public enum NestedAttributes: NestedSpanAttributesProtocol {
+        case namespace
+        public var statusCode: SpanAttributeKey<Int> {
+            "http.status_code"
+        }
+    }
+
+    public subscript<T>(dynamicMember dynamicMember: KeyPath<NestedAttributes, SpanAttributeKey<T>>) -> SpanAttribute? {
```
ktoso, line 191: I THINK this can be removed or moved into SpanAttributeNamespace btw...

PR #106, opened 2020-08-03
Export OpenTelemetryInstrumentationSupport library 📦

slashmo:
`OpenTelemetryInstrumentationSupport` cannot be used because it's not defined as a product in the package file. This PR changes that.


ktoso (commented):
👍

PR #107, opened 2020-08-04
[Wording] Use context instead of baggage

slashmo:


Sources/TracingInstrumentation/Span.swift
```diff
@@ -35,7 +35,7 @@ public protocol Span {
     var endTimestamp: Timestamp? { get }
 
     /// The read-only `BaggageContext` of this `Span`, set when starting this `Span`.
-    var baggage: BaggageContext { get }
+    var context: BaggageContext { get }
```
ktoso, line 38: 👍

ktoso (approved):
Very good, thanks -- we're aligned with our actual naming guidelines now :)

PR #109, opened 2020-08-04
Fix NIO & HTTP Client version clash in UseCases 📦

slashmo:
Having both `NIO` & `AsyncHTTPClient` as explicit dependencies made the resolution fail.

### DIFF
--- a/UseCases/Package.swift
+++ b/UseCases/Package.swift
@@ -11,7 +11,7 @@ let package = Package(
     dependencies: [
         .package(name: "gsoc-swift-tracing", path: "../"),
         .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.1.1"),
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.18.0")
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.9.0"),
     ],
     targets: [
         .target(name: "ManualContextPropagation", dependencies: [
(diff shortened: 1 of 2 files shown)


slashmo:
E.g. here: https://github.com/slashmo/gsoc-swift-tracing/runs/945048329?check_suite_focus=true

ktoso:
That's very weird... I've asked the team for feedback why the error message is bad like that:

```
Fetching https://github.com/slashmo/gsoc-swift-baggage-context
Fetching https://github.com/apple/swift-nio.git
Fetching https://github.com/apple/swift-nio-ssl.git
Fetching https://github.com/apple/swift-log.git
error: Failed to clone https://github.com/apple/swift-nio.git:
    Cloning into bare repository '/__w/gsoc-swift-tracing/gsoc-swift-tracing/UseCases/.build/repositories/swift-nio-b68c973e'...
    fatal: unable to access 'https://github.com/apple/swift-nio.git/': transfer closed with outstanding read data remaining
Fetching https://github.com/apple/swift-nio-extras.git
##[error]Process completed with exit code 1.
```

While it seems more like it should be a version conflict?

UseCases/Package.swift
```diff
@@ -10,8 +10,7 @@ let package = Package(
     ],
     dependencies: [
         .package(name: "gsoc-swift-tracing", path: "../"),
-        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.1.1"),
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.18.0")
+        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.1.1")
```
ktoso: 1) would it work if you depended on httpclient 1.6.0 and nio 2.9?

ktoso: 2. nitpick, please keep the trailing `,` in `[]` listings like here

slashmo: There's no 1.6.0 of `AsyncHTTPClient` 🤔 Latest is 1.1.1: https://github.com/swift-server/async-http-client/releases

slashmo: Using NIO 2.9 & HTTP Client 1.1.1 also fixes the issue. What solution would you prefer?

ktoso: Let's include specific versions NIO 2.9 & HTTP Client 1.1.1 is ok.

ktoso:
I could not reproduce the dependency error...?

Can you confirm this happened also locally? (I'm trying to report this to the SPM team but we can't reproduce)

ktoso (approved):

ktoso:
Looking at it again...     `fatal: unable to access 'https://github.com/apple/swift-nio.git/': transfer closed with outstanding read data remaining` makes me realize perhaps it was just github having some hiccups 🤔

PR #111, opened 2020-08-05
Add recordError method to Span

### LINKED ISSUE #90: Add error recording API to Span
> To facilitate recording an exception languages SHOULD provide a RecordException convenience method. The signature of the method is to be determined by each language and can be overloaded as appropriate. The method MUST record an exception as an Event with the conventions outlined in the exception semantic conventions document.

According to the Otel spec, exceptions must be recorded as `SpanEvent`: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md

slashmo:
Adds `recordError(type:String?,message:String?)` to `Span` with default implementation to record errors as `SpanEvent`s.

Closes #90

### DIFF
--- a/Sources/TracingInstrumentation/Span.swift
+++ b/Sources/TracingInstrumentation/Span.swift
@@ -41,6 +41,12 @@ public protocol Span {
     /// - Parameter event: The `SpanEvent` to add to this `Span`.
     mutating func addEvent(_ event: SpanEvent)
 
+    /// Record an error of the given type described by the the given message.
+    ///
+    /// - Parameters:
+    ///   - error: The error to be recorded.
+    mutating func recordError(_ error: Error)
+
     /// The attributes describing this `Span`.
     var attributes: SpanAttributes { get set }
 
@@ -83,7 +89,7 @@ extension Span {
 // MARK: Span Event
 
 /// An event that occurred during a `Span`.
-public struct SpanEvent {
+public struct SpanEvent: Equatable {
     /// The human-readable name of this `SpanEvent`.
     public let name: String
 
@@ -299,7 +305,7 @@ extension SpanAttribute: ExpressibleByArrayLiteral {
 
 /// A collection of `SpanAttribute`s.
 @dynamicMemberLookup
-public struct SpanAttributes {
+public struct SpanAttributes: Equatable {
     private var _attributes = [String: SpanAttribute]()
 
     /// Create a set of attributes by wrapping the given dictionary.
--- a/Sources/TracingInstrumentation/NoOpTracingInstrument.swift
+++ b/Sources/TracingInstrumentation/NoOpTracingInstrument.swift
@@ -54,6 +54,8 @@ public struct NoOpTracingInstrument: TracingInstrument {
 
         public mutating func addEvent(_ event: SpanEvent) {}
 
+        public func recordError(_ error: Error) {}
+
         public var attributes: SpanAttributes {
             get {
                 [:]
--- a/Tests/TracingInstrumentationTests/TracedLockTests.swift
+++ b/Tests/TracingInstrumentationTests/TracedLockTests.swift
@@ -131,6 +131,8 @@ private final class TracedLockPrintlnTracer: TracingInstrument {
             self.events.append(event)
         }
 
+        func recordError(_ error: Error) {}
+
         mutating func end(at timestamp: Timestamp) {
             self.endTimestamp = timestamp
             print("     span [\(self.operationName): \(self.context[TaskIDKey.self] ?? "no-name")] @ \(timestamp): end")
--- a/Tests/TracingInstrumentationTests/TracingInstrumentTests.swift
+++ b/Tests/TracingInstrumentationTests/TracingInstrumentTests.swift
@@ -168,6 +168,8 @@ struct TestSpan: Span {
         self.events.append(event)
     }
 
+    func recordError(_ error: Error) {}
+
     mutating func end(at timestamp: Timestamp) {
         self.endTimestamp = timestamp
         self.onEnd(self)


Sources/TracingInstrumentation/Span.swift
```diff
@@ -41,6 +41,12 @@ public protocol Span {
     /// - Parameter event: The `SpanEvent` to add to this `Span`.
     mutating func addEvent(_ event: SpanEvent)
 
+    /// Record an error of the given type described by the the given message.
```
ktoso, line 44:
```suggestion
    /// Record an error of the given type described by the the given message.
    
```

Sources/TracingInstrumentation/Span.swift
```diff
@@ -41,6 +41,12 @@ public protocol Span {
...
+    /// - Parameters:
+    ///   - type: The type of error, e.g. "SomeError.case"
+    ///   - message: A message describing the error, e.g. "Division by zero"
+    mutating func recordError(ofType type: String?, message: String?)
```
ktoso: seems a bit manual, not `recordError(_ error: Error)`?

we can `String(reflecting: type(of: error as Any))` and "\\(error)" record it I think?

The combination of the two usually renders pretty well.

Sources/TracingInstrumentation/Span.swift
```diff
@@ -111,6 +129,12 @@ extension SpanEvent: ExpressibleByStringLiteral {
     }
 }
 
+extension SpanEvent: Equatable {
+    public static func == (lhs: SpanEvent, rhs: SpanEvent) -> Bool {
+        lhs.attributes == rhs.attributes
+    }
```
ktoso: this impl seems wrong, why not over all fields?

Sources/TracingInstrumentation/Span.swift
```diff
@@ -77,6 +83,18 @@ extension Span {
...
     }
+
+    public mutating func recordError(ofType type: String?, message: String?) {
+        guard type != nil || message != nil else { return }
+        var attributes = SpanAttributes()
+        if let type = type {
+            attributes["exception.type"] = .string(type)
+        }
+        if let message = message {
+            attributes["exception.message"] = .string(message)
```
ktoso: Makes me think we should NOT implement this to be honest.

Specifically because maybe you're in a system that expects exception, but maybe you're in a system that is more swifty and expects `error.type`.

Swift is not really like the other languages in here: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md as in that errors are not really exceptions, we dont have stacktraces etc...

Makes me feel we should not make that decision; a recorder/tracer should, and it can allow configuring "record errors as "error." or "exception." etc"

WDYT?

ktoso:
It would be actually interesting to ask on the opentelemetry spec, you could ask there for advice what we should do -- `Error` in swift are just values, we do not have stacktraces for them and generally they're called errors -- should we anyway record them as exceptions for nicer interop with the rest of the ecosystem?

Sources/TracingInstrumentation/Span.swift
slashmo: Yep, sounds reasonable 👍

Sources/TracingInstrumentation/Span.swift
slashmo: Not sure what you mean by combining them? Use `String(reflecting:...)` for type and `"\(error)"` for message? Or combine both values to create the kind?

ktoso: Just in the sense that if looking at this event in the backend and I have the type and the string repr, I have the full picture -- i.e. that reporting those two should be enough. But seems we're not in the field of making this call anyway, as the actual reporter should :)

Sources/TracingInstrumentation/Span.swift
ktoso: Ok, let's remove the impl then thx

Sources/TracingInstrumentation/Span.swift
slashmo: Yep, we'd still kind off have to implement it for the test `OTelSpan`, even though it's not of interest to us. I created #112 to rename it.

ktoso: Yup; ~btw maybe good to rename that one to some ExampleSpan? All of our Span are pretty much otel based in a way 🤔~ I should read the entire post before replying :) Yes, sounds good!

Sources/TracingInstrumentation/Span.swift
```diff
@@ -299,7 +323,7 @@ extension SpanAttribute: ExpressibleByArrayLiteral {
 
 /// A collection of `SpanAttribute`s.
 @dynamicMemberLookup
-public struct SpanAttributes {
+public struct SpanAttributes: Equatable {
```
ktoso, line 308: ok

Tests/TracingInstrumentationTests/SpanTests.swift
```diff
@@ -183,6 +183,36 @@ final class SpanTests: XCTestCase {
         XCTAssertEqual(child.links[0].context[TestBaggageContextKey.self], "test")
         XCTAssertEqual(child.links[0].attributes.sampleHttp.statusCode, 418)
     }
+
+    func testRecordErrorDefaultsToCreatingSpanEvent() {
+        var span = OTelSpan(
```
ktoso: the rename?

Sources/TracingInstrumentation/Span.swift
```diff
@@ -42,10 +42,12 @@ public protocol Span {
     mutating func addEvent(_ event: SpanEvent)
 
     /// Record an error of the given type described by the the given message.
+    ///
+    /// It's up to the individual `TracingInstrument` to handle this.
```
ktoso: no need for this comment I think? 

Since we dont provide an impl, it is known that someone has to implement and it's up to it to defined how they handle it.

ktoso (approved):
LGTM 👍

Tests/TracingInstrumentationTests/SpanTests.swift
slashmo: Rebased and force-pushed. I guess you were looking at an outdated version.

Sources/TracingInstrumentation/Span.swift
slashmo: Okay, I added it because I thought some users might have an assumption about what happens here. But I agree that using a protocol communicates that to the users already anyways.

slashmo:
```suggestion
```

PR #113, opened 2020-08-06
Adopt BaggageContext 0.2.0 📦

slashmo:
I updated `BaggageContext` 0.2.0 and made the minimal changes necessary to get everything building. I'm open to spending more time on this to really make use of the additions in `BaggageContext`.


Package.swift
```diff
@@ -14,43 +13,26 @@ let package = Package(
...
-        ),
-
         // ==== --------------------------------------------------------------------------------------------------------
         // MARK: Instrumentation
 
         .target(
             name: "Instrumentation",
             dependencies: [
                 .product(name: "Baggage", package: "swift-baggage-context"),
+                .product(name: "BaggageLogging", package: "swift-baggage-context"),
```
ktoso: does it have to?

Sources/Instrumentation/Instrument.swift
```diff
@@ -50,7 +50,11 @@ public protocol Instrument {
     ///   - carrier: The `Carrier` that was used to propagate values across boundaries.
     ///   - context: The `BaggageContext` into which these values should be injected.
     ///   - extractor: The `Extractor` that extracts values from the given `Carrier`.
-    func extract<Carrier, Extractor>(_ carrier: Carrier, into context: inout BaggageContext, using extractor: Extractor)
+    func extract<Carrier, Extractor>(
+        _ carrier: Carrier,
+        into context: inout BaggageContextProtocol,
```
ktoso: Mixed feelings on the protocol here tbh.

This is a _low level_ API, no need for convenience here IMHO...
Without it would the following work fine?

```
var f : MyFrameworkContext
extract(bla, into: &f.baggage)
```

it would right, so not much difference really
?

Package.swift
slashmo: This was the only way I knew of to get `BaggageLogging` available in the `UseCases` executables.

Sources/TracingInstrumentation/TracingInstrument.swift
```diff
@@ -25,7 +25,7 @@ public protocol TracingInstrument: Instrument {
     ///   - timestamp: The `DispatchTime` at which to start the new `Span`. Passing `nil` leaves the timestamp capture to the underlying tracing instrument.
     func startSpan(
         named operationName: String,
-        context: BaggageContext,
+        context: BaggageContextProtocol,
```
ktoso: 👍 startSpan _is_ end user visible API, so here the protocol makes sense IMO.

UseCases/Sources/HTTPEndToEnd/Services/StorageServiceHandler.swift
```diff
@@ -33,14 +32,15 @@ final class StorageServiceHandler: ChannelInboundHandler {
     func channelRead(context: ChannelHandlerContext, data: NIOAny) {
         guard case .head(let requestHead) = self.unwrapInboundIn(data) else { return }
 
-        var baggage = BaggageContext()
-        baggage[BaggageContext.BaseLoggerKey.self] = self.logger
+        var baggage: BaggageContextProtocol = BaggageContext()
```
ktoso: why the protocol here?

Sources/Instrumentation/Instrument.swift
slashmo: Yep, correct. I'll change it back to `BaggageContext` and test how it affects interacting with the API.

Sources/NIOInstrumentation/BaggageContextInboundHTTPHandler.swift
```diff
@@ -21,16 +21,16 @@ public final class BaggageContextInboundHTTPHandler: ChannelInboundHandler {
     public typealias InboundOut = HTTPServerRequestPart
 
     private let instrument: Instrument
-    private var onBaggageExtracted: (BaggageContext) -> Void
+    private var onBaggageExtracted: (BaggageContextProtocol) -> Void
```
ktoso: Why the protocol? We always deal with the explicit baggage here I think, we should not encourage smuggling around a different type rather than the baggage in the channels as well bts so in NIO the stored thing should be the explicit BaggageContext as well I think

Package.swift
ktoso: Should be able to add it in the UseCases/Package.swift :)

Sources/TracingInstrumentation/TracingInstrument.swift
slashmo: Updated it to accept a `BaggageContextCarrier` instead.

Package.swift
slashmo: How would you do that without adding a dependency on `baggage-context`? None of the following works:

- `.product(name: "BaggageLogging", package: "gsoc-swift-tracing"), // product not found`
- `.product(name: "BaggageLogging", package: "swift-baggage-context"), // unknown package`
- `.product(name: "BaggageLogging", package: "gsoc-swift-baggage-context"), // unknown package`

ktoso: Shared on chat what should work fine

slashmo: 👍

UseCases/Sources/HTTPEndToEnd/Services/StorageServiceHandler.swift
```diff
@@ -34,13 +33,14 @@ final class StorageServiceHandler: ChannelInboundHandler {
         guard case .head(let requestHead) = self.unwrapInboundIn(data) else { return }
 
         var baggage = BaggageContext()
-        baggage[BaggageContext.BaseLoggerKey.self] = self.logger
         self.instrument.extract(requestHead.headers, into: &baggage, using: HTTPHeadersExtractor())
 
-        baggage.logger.info("📦 Looking for the product")
+//        baggage.logger.info("📦 Looking for the product")
+        self.logger.info("📦 Looking for the product")
```
ktoso: mini cleanup, remove the comment?

UseCases/Sources/HTTPEndToEnd/Services/OrderServiceHandler.swift
```diff
@@ -37,14 +36,15 @@ final class OrderServiceHandler: ChannelInboundHandler {
         guard case .head(let requestHead) = self.unwrapInboundIn(data) else { return }
 
         var baggage = BaggageContext()
-        baggage[BaggageContext.BaseLoggerKey.self] = self.logger
 
         self.instrument.extract(requestHead.headers, into: &baggage, using: HTTPHeadersExtractor())
 
-        baggage.logger.info("🧾 Received order request")
+//        baggage.logger.info("🧾 Received order request")
```
ktoso: remove the comment?

UseCases/Sources/HTTPEndToEnd/InstrumentedHTTPClient.swift
```diff
@@ -41,7 +41,7 @@ struct InstrumentedHTTPClient {
     func execute(request: HTTPClient.Request, context: BaggageContext) -> EventLoopFuture<HTTPClient.Response> {
         var request = request
         self.instrument.inject(context, into: &request.headers, using: HTTPHeadersInjector())
-        context.logger.info("🌎 InstrumentedHTTPClient: Execute request")
+//        context.logger.info("🌎 InstrumentedHTTPClient: Execute request")
```
ktoso: should be possible to work now, uncomment?

UseCases/Package.swift
```diff
@@ -13,26 +13,35 @@ let package = Package(
...
         ]),
         .target(name: "HTTPEndToEnd", dependencies: [
-            .product(name: "BaggageLogging", package: "gsoc-swift-tracing"),
-            .product(name: "Instrumentation", package: "gsoc-swift-tracing"),
+            .product(name: "TracingInstrumentation", package: "gsoc-swift-tracing"),
             .product(name: "NIOInstrumentation", package: "gsoc-swift-tracing"),
             .product(name: "AsyncHTTPClient", package: "async-http-client"),
             .product(name: "NIO", package: "swift-nio"),
+            .product(name: "Baggage", package: "swift-baggage-context"),
+            .product(name: "BaggageLogging", package: "swift-baggage-context"),
```
ktoso, line 40: good :)

ktoso (approved):
LGTM, minor nitpick. The usage of the carrier protocol looking good — just in the Span APIs.

UseCases/Sources/HTTPEndToEnd/InstrumentedHTTPClient.swift
slashmo: How? Wouldn't `context` have to be a `LoggingBaggageContextCarrier` in order for that to work?

ktoso: Ah, that's what I get for reviewing from a phone 😉 

Right that's good -- can you remove the comment then though?

UseCases/Sources/HTTPEndToEnd/Services/OrderServiceHandler.swift
```diff
@@ -37,14 +36,15 @@ final class OrderServiceHandler: ChannelInboundHandler {
...
+        self.logger.info("🧾 Received order request")
```
ktoso: this could do `self.logger.with(context).info` if it wanted to

UseCases/Sources/HTTPEndToEnd/InstrumentedHTTPClient.swift
slashmo: Replaced it with a `self.logger` so that the example at least continues to log.

UseCases/Sources/HTTPEndToEnd/InstrumentedHTTPClient.swift
```diff
@@ -41,7 +43,7 @@ struct InstrumentedHTTPClient {
     func execute(request: HTTPClient.Request, context: BaggageContext) -> EventLoopFuture<HTTPClient.Response> {
         var request = request
         self.instrument.inject(context, into: &request.headers, using: HTTPHeadersInjector())
-//        context.logger.info("🌎 InstrumentedHTTPClient: Execute request")
+        self.logger.with(context: context).info("🌎 InstrumentedHTTPClient: Execute request")
```
ktoso, line 46: yup, good

ktoso: (just for the example, a real instrumented one would not log here)

ktoso (approved):
LGTM :)

PR #115, opened 2020-08-11
Add forceFlush method to TracingInstrument

### LINKED ISSUE #85: Add forceFlush to trace instruments
This is a follow up of https://github.com/pokryfka/aws-xray-sdk-swift/pull/16#discussion_r457828748.

I do not have a strong opinion on that, still I think its worth a discussion and it exceeds scope of a comment in PR.

---

[`XRayRecorder`](https://github.com/pokryfka/aws-xray-sdk-swift/blob/master/Sources/AWSXRayRecorder/Recorder.swift) currently exposes:

the recorder currently exposes "NIO" flush:

```swift
    public func flush(on eventLoop: EventLoop) -> EventLoopFuture<Void> {}
```

as well as blocking wait without NIO dependency:

```swift
    public func wait(_ callback: ((Error?) -> Void)? = nil) { // }
```

@ktoso

> Could you handle this in the recorder's shutdown?

yes for "services" (Vapor, Kitura, ...)

however in case of "lambda runtime", my understanding is that we do need to emit all the segments during event handling (event though it will make the request handling longer) because otherwise lambda may be suspended (and potentially never wake up again) before emitting finishes, example:

```swift
private struct ExampleLambdaHandler: EventLoopLambdaHandler {
    typealias In = Cloudwatch.ScheduledEvent
    typealias Out = Void

    private let recorder = XRayRecorder()

    private func doWork(on eventLoop: EventLoop) -> EventLoopFuture<Void> {
        eventLoop.submit { usleep(100_000) }.map { _ in }
    }

    func handle(context: Lambda.Context, event: In) -> EventLoopFuture<Void> {
        recorder.segment(name: "ExampleLambdaHandler", context: context) {
            self.doWork(on: context.eventLoop)
        }.flatMap {
            self.recorder.flush(on: context.eventLoop)
        }
    }
}

Lambda.run(ExampleLambdaHandler())
```

to be honest, while my explanation above does make sense to me, I do plan checking more in details how its implemented/enforced in "official" XRay SDKs;

also note that while in case of XRay emitting = sending UDP in local network, which should be comparatively inexpensive and may be finished (subject of more testing) before lambda is suspended even without explicit flushing;
emitting spans using different `TracingInstrument` in lambda handler will be typically much mroe expensive and will take more time;

slashmo:
PRified @ktoso's suggestion.
Closes #85 

CC @pokryfka

### DIFF
--- a/Sources/TracingInstrumentation/NoOpTracingInstrument.swift
+++ b/Sources/TracingInstrumentation/NoOpTracingInstrument.swift
@@ -25,6 +25,8 @@ public struct NoOpTracingInstrument: TracingInstrument {
         NoOpSpan()
     }
 
+    public func forceFlush() {}
+
     public func inject<Carrier, Injector>(
         _ context: BaggageContext,
         into carrier: inout Carrier,
--- a/Sources/TracingInstrumentation/TracingInstrument.swift
+++ b/Sources/TracingInstrumentation/TracingInstrument.swift
@@ -29,6 +29,14 @@ public protocol TracingInstrument: Instrument {
         ofKind kind: SpanKind,
         at timestamp: Timestamp?
     ) -> Span
+
+    /// Export all ended spans to the configured backend that have not yet been exported.
+    ///
+    /// This function should only be called in cases where it is absolutely necessary,
+    /// such as when using some FaaS providers that may suspend the process after an invocation, but before the backend exports the completed spans.
+    ///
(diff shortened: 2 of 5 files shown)


pokryfka:
awesome, thank you

ktoso (approved):
LGTM, what about shutdown?

slashmo:
> LGTM, what about shutdown?
@ktoso 

`shutdown` will land in a separate PR, tracked in #116

PR #120, opened 2020-08-12
Tweak naming of addLink convenience

### LINKED ISSUE #117: Revisit addLink parameter wording (it should not be `parent`?)
The span passed to add link does not have to necessarily be a parent;

> When using the scatter/gather (also called fork/join) pattern, the root operation starts multiple downstream processing operations and all of them are aggregated back in a single Span. This last Span is linked to many operations it aggregates.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/overview.md#links-between-spans

today we have it as `public mutating func addLink(_ parent: Span, attributes: SpanAttributes = [:]) {` it should be `_ other` I guess, parent implies too much here. It may or may not be a parent.

slashmo:
Closes #117

### DIFF
--- a/Sources/TracingInstrumentation/Span.swift
+++ b/Sources/TracingInstrumentation/Span.swift
@@ -77,11 +77,11 @@ extension Span {
         self.end(at: .now())
     }
 
-    /// Adds a `SpanLink` to the given parent `Span`.
-    /// - Parameter parent: The parent `Span`.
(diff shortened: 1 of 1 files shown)


ktoso (approved):
LGTM.

I always wonder how much of the spec we should paste here, but for now let's not at all, this looks good 👍

PR #124, opened 2020-08-13
Use correct defaults in startSpan convenience

slashmo:
They were documented correctly but used the wrong defaults.


ktoso (commented):
Late LGTM

PR #128, opened 2020-08-17
Downgrade to Swift 5.0

slashmo:
Adds support for Swift 5.0 & Swift 5.1 which is necessary for being integrated by other libraries that still support these versions. The CI now executes tests in parallel on 5.0.3, 5.1, and 5.2.


Package.swift
```diff
@@ -73,22 +72,22 @@ let package = Package(
...
+                "OpenTelemetryInstrumentationSupport",
             ]
         ),
 
         // ==== --------------------------------------------------------------------------------------------------------
         // MARK: Performance / Benchmarks
 
         .target(
-            name: "Benchmarks",
+            name: "TracingBenchmarks",
```
slashmo, line 83: Renamed both `Benchmarks` & `SwiftBenchmarkTools` to avoid clashing with the ones defined in `swift-baggage-context`.

ktoso: Good, we should do a similar boring prefix rename in there as well I guess 👍

Package.swift
```diff
@@ -37,23 +36,23 @@ let package = Package(
         .target(
             name: "TracingInstrumentation",
             dependencies: [
-                "Instrumentation"
+                "Instrumentation",
```
ktoso, line 39: thx :)

Package.swift
slashmo: Agreed, let's rename them to `BaggageBenchmarks` & `BaggageBenchmarkTools`. WDYT?

Sources/TracingInstrumentation/Span.swift
```diff
@@ -190,9 +192,11 @@ public enum SpanAttribute: Equatable {
     case array([SpanAttribute])
     case stringConvertible(CustomStringConvertible)
 
+    #if compiler(>=5.2)
     /// This is a "magic value" that is used to enable the KeyPath based accessors to specific attributes.
     /// This value will never be stored or returned, and any attempt of doing so would WILL crash your application.
     case __namespace
+    #endif
```
ktoso: not sure we should do this about the enum case; keep it in, even if we dont use it.

Also, we should see if we can remove this __namespace -- it's a bit of a hack.
Can you make a ticket for it? "See if we can remove the case __namespace in attributes"?

Sources/TracingInstrumentation/Span.swift
```diff
@@ -353,7 +363,33 @@ public struct SpanAttributes: Equatable {
         where Namespace: SpanAttributeNamespace {
         SpanAttribute.__namespace[keyPath: dynamicMember]
     }
+}
+#else
```
ktoso: why the entire type? can we not #if only the "offending" subscript?

Tests/TracingInstrumentationTests/SpanTests.swift
```diff
@@ -128,9 +129,11 @@ final class SpanTests: XCTestCase {
         XCTAssertEqual(attributes.name, SpanAttribute.string("kappa"))
         XCTAssertEqual(attributes.name, "kappa")
         XCTAssertEqual(attributes.sampleHttp.statusCode, 200)
+        #endif
```
ktoso, line 133:
```
#else 
print some info about shy this is skipped?
#endif
```

Tests/TracingInstrumentationTests/SpanTests.swift
```diff
@@ -175,13 +179,19 @@ final class SpanTests: XCTestCase {
             onEnd: { _ in }
         )
 
+        #if compiler(>=5.2)
         var attributes = SpanAttributes()
         attributes.sampleHttp.statusCode = 418
+        #else
+        let attributes = SpanAttributes()
```
ktoso: the corresponding not typesafe fancy version here?

```
attributes["http.status_code"] = .int(418)
```

would it be?

UseCases/Package.swift
```diff
@@ -10,38 +10,37 @@ let package = Package(
...
+            "NIO",
+            "Baggage",
+            "BaggageLogging",
         ]),
         .target(name: "InstrumentsAppTracing", dependencies: [
-            .product(name: "Instrumentation", package: "gsoc-swift-tracing"),
-            .product(name: "TracingInstrumentation", package: "gsoc-swift-tracing"),
+            "Instrumentation",
+            "TracingInstrumentation",
         ]),
```
ktoso, line 44: looks good

scripts/sanity.sh
```diff
@@ -15,6 +15,18 @@
...
+ruby "$here/../scripts/generate_linux_tests.rb" > /dev/null
+SECOND_OUT="$(git status --porcelain)"
+if [[ "$FIRST_OUT" != "$SECOND_OUT" ]]; then
+  printf "\033[0;31mmissing changes!\033[0m\n"
+  git --no-pager diff
+  exit 1
+else
+  printf "\033[0;32mokay.\033[0m\n"
+fi
+
```
ktoso, line 29: good thanks!

ktoso (commented):
Some comments in-line, shaping up well. 

We should also document this tradeoff but perhaps we can do a big documentation day one day soon... we should write up a "pitch" and overall design info in README -- we can do that as one big task soon

Package.swift
ktoso: Sounds good

slashmo: https://github.com/slashmo/gsoc-swift-baggage-context/pull/21

Sources/TracingInstrumentation/Span.swift
slashmo: https://github.com/slashmo/gsoc-swift-tracing/issues/129

slashmo: I left this out on purpose because it's not being used anywhere but the 5.2 "magic". What would be the benefit of leaving it in for < 5.2?

Tests/TracingInstrumentationTests/SpanTests.swift
slashmo: What do you mean by "print"? Something like `#warning` or an actual runtime `print`, or simply a comment?

Sources/TracingInstrumentation/Span.swift
slashmo: It's slightly more than just the subscript. Prior to 5.2 we e.g. won't need `@dynamicMemberLookup` on `SpanAttributes`, right? Also, `subscript(_ name: String)` in the case of < 5.2 doesn't need to handle `__namespace`.

Tests/TracingInstrumentationTests/SpanTests.swift
slashmo: Added ✅

Tests/TracingInstrumentationTests/SpanTests.swift
slashmo: Added a comment for it.

slashmo:
@ktoso Feel free to *squash* this one if accepted. Had some smaller commits fixing up unit tests / CI.

Sources/TracingInstrumentation/Span.swift
ktoso: If you have a library that supports all swift versions, and they happen to switch over the attributes enum, then they'd be unable to write ONE switch implementation that works on all compiler versions without using `default` since they have to handle `__namespace` on 5.2+ and they couldn't handle it on previous versions because it's compiled out.

We should keep enums always the same across all supported compiler versions

Tests/TracingInstrumentationTests/SpanTests.swift
ktoso: I meant `print("oh oh")` to be honest :) 
In my own test suites I usually print something in yellow for such tests...

but either's fine really

ktoso:
This is LGTM once the enum __namespace is not excluded 👍

PR #131, opened 2020-08-19
Add (g)RPC semantic attributes

slashmo:
Adds an `rpc` namespace with the semantic span attributes as defined in the OTel Spec: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#events

### DIFF
--- a/Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+GRPCSemantics.swift
+++ b/Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+GRPCSemantics.swift
@@ -1,0 +1,71 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Tracing open source project
+//
+// Copyright (c) 2020 Moritz Lang and the Swift Tracing project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import TracingInstrumentation
+
+extension SpanAttributeName {
+    /// - See: GRPCAttributes
+    public static let gRPCMessageType = "message.type"
+    /// - See: GRPCAttributes
+    public static let gRPCMessageID = "message.id"
+    /// - See: GRPCAttributes
+    public static let gRPCMessageCompressedSize = "message.compressed_size"
+    /// - See: GRPCAttributes
+    public static let gRPCMessageUncompressedSize = "message.uncompressed_size"
+}
+
+#if swift(>=5.2)
+extension SpanAttributes {
+    /// Semantic conventions for gRPC spans.
+    public var gRPC: GRPCAttributes {
+        get {
+            .init(attributes: self)
+        }
+        set {
+            self = newValue.attributes
+        }
+    }
+}
+
+/// Semantic conventions for gRPC spans as defined in the OpenTelemetry spec.
+///
+/// - SeeAlso: [OpenTelemetry: Semantic conventions for gRPC spans](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#grpc) (as of August 2020)
+@dynamicMemberLookup
+public struct GRPCAttributes: SpanAttributeNamespace {
+    public var attributes: SpanAttributes
+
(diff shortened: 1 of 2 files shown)


slashmo:
@ktoso & @glbrntt I'm wondering whether we should type out the message attributes as well: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#events

Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+RPCSemantics.swift
```diff
@@ -0,0 +1,39 @@
...
+// See LICENSE.txt for license information
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=5.2)
+import TracingInstrumentation
+
+/// Semantic conventions for HTTP spans as defined in the OpenTelemetry spec.
```
ktoso:
```suggestion
/// Semantic conventions for RPC span attributes as defined in the OpenTelemetry spec.
```

ktoso: Can we add a 

```
/// - SeeAlso: [OpenTelemetry: Semantic conventions for RPC spans](https://github.com/open-telemetry/opentelemetry-specification/blob/b70565d5a8a13d26c91fb692879dc874d22c3ac8/specification/trace/semantic_conventions/rpc.md) (as of August 2020)
```

ktoso:
> @ktoso & @glbrntt I'm wondering whether we should type out the message attributes as well: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#events

Yeah might as well, seems good

Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+RPCSemantics.swift
```diff
@@ -0,0 +1,39 @@
...
+//
+// See LICENSE.txt for license information
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=5.2)
+import TracingInstrumentation
+
```
ktoso, line 15: Also the namespace var:

```suggestion
extension SpanAttributes {
    /// Semantic conventions for RPC spans.
    public var rpc: RPCAttributes {
        get {
            .init(attributes: self)
        }
        set {
            self = newValue.attributes
        }
    }
}

```

Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+RPCSemantics.swift
slashmo: Yeah, that's a good addition. I'll add it to the other ones as well 👍

slashmo:
> > @ktoso & @glbrntt I'm wondering whether we should type out the message attributes as well: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#events
> 
> Yeah might as well, seems good

Implemented in the latest commit.

Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+GRPCSemantics.swift
```diff
@@ -0,0 +1,71 @@
...
+// See LICENSE.txt for license information
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import TracingInstrumentation
+
+extension SpanAttributeName {
+    /// - See: GRPCAttributes
```
ktoso, line 17: I think those need to be in `` for jazzy to make it into links?

We can do a proper documentation pass over all docs next week though.

ktoso (approved):
👍

Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+GRPCSemantics.swift
slashmo: Yeah, let's do it next week when we go through everything 👀

PR #133, opened 2020-08-19
Add SpanAttributeName String constants

### LINKED ISSUE #132: Backwards-compatible SpanAttribute string constants
Since we added Swift >= 5.0 support our semantic `Span` attributes (which are >= 5.2-only) cannot be used in libraries that make use of this backward-compatibility. Therefore, we want to expose string constants for the attribute keys that can be used with any Swift version. We'll then reuse these constants when defining the 5.2-only type-safe attributes.

slashmo:
Closes #132

### DIFF
--- a/Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+EndUser.swift
+++ b/Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+EndUser.swift
@@ -14,8 +14,11 @@
 import TracingInstrumentation
 
 extension SpanAttributeName {
+    /// - See: EndUserAttributes
     public static let endUserID = "enduser.id"
+    /// - See: EndUserAttributes
     public static let endUserRole = "enduser.role"
+    /// - See: EndUserAttributes
     public static let endUserScope = "enduser.scope"
 }
 
--- a/Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+EndUser.swift
+++ b/Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+EndUser.swift
@@ -11,9 +11,15 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if swift(>=5.2)
 import TracingInstrumentation
 
+extension SpanAttributeName {
+    public static let endUserID = "enduser.id"
+    public static let endUserRole = "enduser.role"
+    public static let endUserScope = "enduser.scope"
+}
+
+#if swift(>=5.2)
 extension SpanAttributes {
     /// Semantic end-user attributes.
     public var endUser: EndUserAttributes {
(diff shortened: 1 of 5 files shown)


Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+EndUser.swift
```diff
@@ -33,6 +33,8 @@ extension SpanAttributes {
 }
 
 /// End-user-related semantic conventions as defined in the OpenTelemetry spec.
+///
+/// - SeeAlso: [OpenTelemetry: General identity attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/b70565d5a8a13d26c91fb692879dc874d22c3ac8/specification/trace/semantic_conventions/span-general.md#general-identity-attributes) (as of August 2020)
```
ktoso, line 40: cool good 👍

Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+EndUser.swift
```diff
@@ -39,14 +47,14 @@ public struct EndUserAttributes: SpanAttributeNamespace {
         public init() {}
 
         /// Username or client_id extracted from the access token or Authorization header in the inbound request from outside the system.
-        public var id: SpanAttributeKey<String> { "enduser.id" }
+        public var id: SpanAttributeKey<String> { .init(name: SpanAttributeName.endUserID) }
 
         /// Actual/assumed role the client is making the request under extracted from token or application security context.
-        public var role: SpanAttributeKey<String> { "enduser.role" }
+        public var role: SpanAttributeKey<String> { .init(name: SpanAttributeName.endUserRole) }
```
ktoso, line 56: meh yeah too bad we mis the expressible by...

it's fine tho, good 👍

Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+EndUser.swift
```diff
@@ -11,9 +11,15 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if swift(>=5.2)
 import TracingInstrumentation
 
+extension SpanAttributeName {
+    public static let endUserID = "enduser.id"
+    public static let endUserRole = "enduser.role"
+    public static let endUserScope = "enduser.scope"
```
ktoso, line 22: I wonder if we should 

/// - See: `EndUserAttributes`

on all of those to at least link like that; I don't want to copy all of the documentation here though, but at least in the docs rendering each attr can point at the right struct to look for full docs.

Sources/OpenTelemetryInstrumentationSupport/SpanAttributeName.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+// See LICENSE.txt for license information
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import TracingInstrumentation
+
+/// A set of `String` constants describing semantic names for `SpanAttributes`.
```
ktoso:
```suggestion
/// Namespace for attribute key constants used with `SpanAttributes`.
```

?

ktoso (commented):
Minor comments, looks good

Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+EndUser.swift
slashmo: Yep, that's a nice compromise.

ktoso (approved):

PR #136, opened 2020-08-20
Namespace SpanAttributeNames

### LINKED ISSUE #135: Consider grouping span attribute names in namespaces
Just an idea...

I think this would compose better (plus much nicer generated documentation and autocompletion support):

```swift
extension SpanAttributeName {
    /// Semantic conventions for HTTP spans.
    public enum HTTP {
        public static let method = "http.method"
        public static let url = "http.url"
    }
}
```

than

```swift
extension SpanAttributeName {
    public static let httpMethod = "http.method"
    public static let httpURL = "http.url"
}
```

slashmo:
Closes #135

### DIFF
--- a/Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+EndUser.swift
+++ b/Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+EndUser.swift
@@ -15,11 +15,14 @@ import TracingInstrumentation
 
 extension SpanAttributeName {
     /// - See: EndUserAttributes
-    public static let endUserID = "enduser.id"
-    /// - See: EndUserAttributes
-    public static let endUserRole = "enduser.role"
-    /// - See: EndUserAttributes
-    public static let endUserScope = "enduser.scope"
+    public enum EndUser {
+        /// - See: EndUserAttributes
+        public static let id = "enduser.id"
+        /// - See: EndUserAttributes
+        public static let role = "enduser.role"
+        /// - See: EndUserAttributes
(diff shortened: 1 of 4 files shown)


Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+EndUser.swift
```diff
@@ -15,11 +15,14 @@ import TracingInstrumentation
 
 extension SpanAttributeName {
     /// - See: EndUserAttributes
-    public static let endUserID = "enduser.id"
-    /// - See: EndUserAttributes
-    public static let endUserRole = "enduser.role"
-    /// - See: EndUserAttributes
-    public static let endUserScope = "enduser.scope"
+    public enum EndUser {
```
ktoso, line 18: `Enduser` since we dont uppercase or separate it in the IDs either? (it is `enduser` not `end_user`)?

ktoso (approved):
LGTM :)

Sources/OpenTelemetryInstrumentationSupport/SpanAttribute+EndUser.swift
slashmo: Mh, was on the fence about this. In the ID we don't have a choice if we want to be Otel-compatible, but grammatically correct would be end-user (at least according to Grammarly 😁).

ktoso: Not a strong opinion, we can go with this 👍

PR #141, opened 2020-08-31
Drop Instrument suffix in TracingInstrument & TracingInstrumentation

### LINKED ISSUE #140: Consider renaming TracingInstrument -> Tracer as it's a well known word
And just "happens to be" an instrument as well.

It causes less typing in instrumented systems / user apps and is more familiar to existing users.
We should not make our API too annoying to use as people won't like it otherwise.

Our Tracer also aligns with OTel so that's good.

slashmo:
Closes #140


ktoso (approved):
LGTM, Let's gooo~

PR #142, opened 2020-08-31
Require Span types to be classes

### LINKED ISSUE #119: Force Spans to be classes?
I remain on the fence about it... So far I've seen a few implementation implementations of tracers and only one of them uses structs for spans, though it's also very simplistic and should really also be a class. (All notable impls, xray and a few others I looked into though cannot share are really best implemented by classes).

Most specifically, this is more about the functions being mutating and that then we have `var span` and we change that, rather than mutating the span which does not feel as natural. Context is different because it's a bag of values and naturally has value semantics, but spans not really -- and spans _do_ want `deinit` because they _do_ want to report any dropped spans.

Specifically in patterns like this:

```
var/let span = tracer.startSpan("...", context: context)

someAsync { 
  someAsync { span.end() } // double ending, bad!
  span.end() // "mutating" because sets flags/timestamps
}
```

the value semantics actually are not so good here, because we'd also want to detect a double `end()`, and open telemetry advices:

> Implementations MUST ignore all subsequent calls to End (there might be exceptions when Tracer is streaming event and has no mutable state associated with the Span).

So yeah, that basically means spans are mutable so might as well embrace it -- forcing all spans to be classes.

Any concerns about this?

slashmo:
Closes #119

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -7,14 +7,14 @@ let package = Package(
         .library(name: "Instrumentation", targets: ["Instrumentation"]),
         .library(name: "Tracing", targets: ["Tracing"]),
         .library(name: "NIOInstrumentation", targets: ["NIOInstrumentation"]),
-        .library(name: "OpenTelemetryInstrumentationSupport", targets: ["OpenTelemetryInstrumentationSupport"])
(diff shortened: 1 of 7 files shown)


ktoso (approved):
LGTM, realistically this seems to be the right thing to do.

We'll play around in more use cases and may have to revisit, but so far I think this is the right thing to do it seems

PR #143, opened 2020-08-31
Store BaggageContext in NoOpSpan

### LINKED ISSUE #138: NoOp Tracing/Span must propagate context
Otherwise we drop baggage which is used to propagate "just logging".

See also https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#behavior-of-the-api-in-the-absence-of-an-installed-sdk

slashmo:
Closes #138


ktoso (approved):
Could get a test?

That without enabling any tracer we still propagate context.

slashmo:
> Could get a test?

Good point, will do 👌

Tests/TracingTests/InstrumentationSystemTests.swift
```diff
@@ -0,0 +1,40 @@
...
+import Tracing
+import XCTest
+
+final class InstrumentationSystemTests: XCTestCase {
+    func testItProvidesAccessToATracer() {
+        let tracer = TestTracer()
+
+        XCTAssertNil(InstrumentationSystem.tracer(of: TestTracer.self))
+
+        InstrumentationSystem.bootstrapInternal(tracer)
```
ktoso: good, was just about to comment on this, previous test version used `bootstrap()` which would have been wrong;

do we also need a way to "unset" it after the test runs or not hmmmm?

slashmo: Yeah, should we just make the `bootstrapInternal` method accept an optional? Or is there an opposite better describing it?

ktoso: accepting optional sounds good to me, good enough for our tests and similar use cases for a "reset" 👍

slashmo: Done ✅

ktoso:
Nice thanks!

PR #146, opened 2020-09-08
Join multiple header values in HTTPHeadersExtractor

### LINKED ISSUE #145: ExtractorProtocol cannot handle multiple header fields
Currently, `ExtractorProtocol` is defined with the following method requirement, which works well enough for single-field headers such as W3C TraceContexts `traceparent`.

```swift
func extract(key: String, from carrier: Carrier) -> String?
```

For `tracestate`, however, W3C TraceContext allows multiple header fields:

> TraceState *MAY* be sent or received as multiple header fields
> https://www.w3.org/TR/trace-context/#tracestate-header-field-values

Therefore, we should adapt our `ExtractorProtocol` to allow the extraction of multiple values for the same key.

In regards to `tracestate` the `InjectorProtocol` doesn't have to be changed:

> When receiving multiple `tracestate` header fields, they *MUST* be combined into a single header according to [RFC7230](https://httpwg.org/specs/rfc7230.html#field.order).
> https://www.w3.org/TR/trace-context/#tracestate-header-field-values

slashmo:
Closes #145


Sources/NIOInstrumentation/HTTPHeadersCarrier.swift
```diff
@@ -18,7 +18,11 @@ public struct HTTPHeadersExtractor: ExtractorProtocol {
     public init() {}
 
     public func extract(key: String, from headers: HTTPHeaders) -> String? {
-        return headers.first(name: key)
+        return headers
+            .lazy
+            .filter { $0.name == key }
+            .map { $0.value }
+            .joined(separator: ",")
```
ktoso, line 30: please document this :) 

With links as `- SeeAlso: ... ` to the spec as we looked into today.

Best on the struct itself.

ktoso (approved):
LGTM, please add docs before merging 👍

ktoso (approved):
LGTM

PR #147, opened 2020-09-29
Return nil when extracting empty header value

slashmo:
Due to the changes introduced in #146, when `HTTPHeadersExtractor` extracts no headers, it returns an empty string instead of `nil`, which may lead to slight misbehavior for users of this API.

### DIFF
--- a/Sources/NIOInstrumentation/HTTPHeadersCarrier.swift
+++ b/Sources/NIOInstrumentation/HTTPHeadersCarrier.swift
@@ -23,11 +23,11 @@ public struct HTTPHeadersExtractor: ExtractorProtocol {
     public init() {}
 
     public func extract(key: String, from headers: HTTPHeaders) -> String? {
-        return headers
+        let headers = headers
             .lazy
             .filter { $0.name == key }
             .map { $0.value }
-            .joined(separator: ",")
+        return headers.isEmpty ? nil : headers.joined(separator: ",")
     }
 }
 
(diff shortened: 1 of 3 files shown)


Sources/NIOInstrumentation/HTTPHeadersCarrier.swift
```diff
@@ -23,11 +23,12 @@ public struct HTTPHeadersExtractor: ExtractorProtocol {
     public init() {}
 
     public func extract(key: String, from headers: HTTPHeaders) -> String? {
-        return headers
+        let header = headers
             .lazy
             .filter { $0.name == key }
             .map { $0.value }
             .joined(separator: ",")
+        return header.isEmpty ? nil : header
```
slashmo: @ktoso I'm wondering whether this check should be on the array instead of the empty string. If we check on the String, users can't distinguish between no header value and an empty but present header value.

ktoso:
Oh great catch! LGTM

ktoso (approved):

Sources/NIOInstrumentation/HTTPHeadersCarrier.swift
ktoso: You may be right here... on the array is more precise. In practice I wonder is empty values are meaningful at all though... is it even legal http to have an empty header, I don't remember tbh.

slashmo: Okay, even if empty header values are illegal/impossible, in my opinion, the extractor shouldn't be the "judge" of it. I'll change it to check on the array.

ktoso: Sounds good 👍

PR #152, opened 2020-10-07
Correctly forward channel events in NIO instrumentation handlers

slashmo:
Due to early-exit in a guard-statement in both NIO instrumentation handlers, I missed forwarding read/write events to chained handlers. This has been fixed and regression-tested now.

I also renamed both handlers to include Server/Client to indicate their purpose more clearly.


ktoso:
Feel free to merge :) sorry I'm slow on reviews this week

ktoso (approved):

